PR #5, opened 2026-01-19
feat: add UI constants and enhance ToolResultBlock

lybtt:
image

优化当前超长文本result的显示效果，动态判断是否折叠显示


hqhq1025:
Hi @lybtt, sorry for the long wait on this — we should have gotten back to you sooner.

The core idea here (auto-collapsing long tool results with a preview) is genuinely useful UX, and we appreciate you putting it together. Unfortunately, `MessageCard.tsx` has gone through quite a few changes since January (33+ commits), and the current `ToolResultBlock` now includes `memo()` wrapping, tool name resolution, deduplication logic, and image support — so this PR would have significant merge conflicts and would regress some of those features.

If you're still interested, we'd love to see a fresh PR based on current `main` that adds the collapse/preview behavior on top of the existing component. A couple of things to keep in mind for the next round:
- Use `t()` for user-facing strings (i18n)
- Keep English for code comments to stay consistent with the rest of the codebase

Thanks again for the contribution, and sorry for the delay!

PR #73, opened 2026-03-22
fix: stabilize Windows MCP packaging and logger tests

Sun-sunshine06:
## Summary
- stage bundled MCP outputs into .bundle-resources/mcp before packaging so electron-builder no longer copies directly from freshly-written dist-mcp files on Windows
- add retry-based MCP staging helpers and a regression test covering the staged resource path
- make logger context tests wait for log files to appear and stabilize before reading to reduce file I/O flakes

## Validation
- npm.cmd run typecheck
- npm.cmd test -- --run tests/logger-context.test.ts tests/bundle-mcp-script.test.ts tests/build-windows-script.test.ts tests/windows-legacy-uninstall-remediation.test.ts
- npm.cmd run build:win

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -93,6 +93,7 @@ DEV_LOG.md
 
 # Build cache
 .build-cache/
+.bundle-resources/
 
 # Temp directories
 .tmp/
--- a/electron-builder.yml
+++ b/electron-builder.yml
@@ -46,11 +46,11 @@ files:
   - "!**/.eslintrc*"
   - "!**/tsconfig.json"
 
-# MCP servers (compiled JavaScript files, cross-platform)
-# NOTE: Platform-specific extraResources OVERRIDE this global section,
-# so dist-mcp must also be listed in each platform's extraResources.
+# MCP servers are staged into .bundle-resources/mcp after build:mcp completes.
+# Packaging from the staged directory is more reliable on Windows than reading
+# directly from freshly emitted dist-mcp files.
 extraResources:
-  - from: dist-mcp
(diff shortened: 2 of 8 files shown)


hqhq1025 (changes requested):
## ⚠️ Request Changes — Critical merge conflict with current main

The MCP staging approach is sound and correctly solves the Windows EBUSY issue.

### 🚨 Blocking: `decryptWithFallback` authTag parameter

This PR modifies `credentials-store.ts`'s `decryptWithFallback` signature, but it appears to be based on an older version of the file. The current `main` branch has `authTag?: string` as a third parameter (line ~212). If this PR merges as-is, the `authTag` parameter would be silently dropped, causing **all GCM-encrypted credentials to fail decryption** (`ERR_CRYPTO_INVALID_TAG`).

**Action needed**: Rebase onto current main and reconcile the `decryptWithFallback` signature.

### Other findings
- **Medium**: `replaceDirectoryWithRetries` silently returns on ambiguous rename failure (source disappears mid-operation). Consider asserting the destination contains expected files before returning.
- **Low**: Dead code in `removePathWithRetries` and `copyFileWithRetries` — unreachable `lastError` throw after the loop
- **Medium**: Logger test stabilization replaces stream `finish` event with polling, which is less reliable
- **Low**: Variable shadowing `logFilePath` in the new `getLogContent`

Sun-sunshine06:
Closing this stale PR because the relevant changes were already cherry-picked into \main\ by the maintainer, per the review notes on April 30, 2026.

PR #107, opened 2026-04-14
fix(remote): sync Feishu DM policy to gateway auth mode

### LINKED ISSUE #92: 连接飞书反复失败
### App Version

3.3.0-beta8

### Platform

Windows

### Bug Description

飞书验证码模式则不规则弹出“⚠️ 您没有权限使用此机器人。请联系管理员获取访问权限。”其他模式则经常漏收消息，或者干脆无法收到消息

### Steps to Reproduce

飞书

### Expected Behaviour

收到消息

### Actual Behaviour

消息失败

```text

```

### Additional Context

_No response_

hqhq1025:
## Summary

Fixes #92 — Feishu authorization failure after restart when DM policy is set to 'open'.

**Root cause:** When a user sets the Feishu DM policy to `open` in the UI, only `channels.feishu` config is updated. But `gateway.auth.mode` stays at the default `allowlist` with an empty allowlist, so `checkAuthorization()` rejects all users.

**Changes:**
- Add `'open'` to `GatewayAuthConfig.mode` type union (`types.ts`)
- Add `case 'open': return true` in `checkAuthorization()` (`gateway.ts`)
- Sync Feishu DM policy to gateway auth mode in `updateFeishuConfig()` (`remote-manager.ts`): maps `open` → `open`, `pairing` → `pairing`, `allowlist` → `allowlist` with allowFrom list

## Test plan

- [ ] Set Feishu DM policy to `open` in UI, restart, verify all users can send messages
- [ ] Set Feishu DM policy to `allowlist` with specific users, verify only those users are authorized
- [ ] Set Feishu DM policy to `pairing`, verify pairing flow works
- [ ] `npx tsc --noEmit` passes with no errors


hqhq1025 (commented):
Re: gateway-wide auth scope concern —

Investigated the actual impact:

1. **Only Feishu is implemented** — no other channel has a runtime implementation, so cross-channel leakage doesn't occur today
2. **WebSocket clients bypass `checkAuthorization()` entirely** — they have their own token-based auth in `handleWSAuth()`, so `open` mode doesn't affect them
3. **Group messages still require `@mention`** — `shouldProcessGroupMessage()` runs after auth and filters non-mentioned messages regardless of auth mode

The concern is architecturally valid for a future multi-channel setup, but not a real issue with the current single-channel codebase. Making auth per-channel would be a significant refactor better suited as a separate issue, not part of this #92 bugfix.

The existing comment at `remote-manager.ts:325-327` already documents this cross-channel scope explicitly.

PR #165, opened 2026-04-27
feat(permissions): gate local-agent tool calls through PermissionDialogMike/feat/permissions

### LINKED ISSUE #164: Local-chat tools execute without approval: `PermissionDialog` and `settings.permissionRules` are not wired to the pi-coding-agent tool path
### App Version

3.3.0

### Platform

macOS

### Bug Description

**Labels:** bug · security · area/permissions
**Affected versions:** open-cowork `3.3.0`, `@mariozechner/pi-coding-agent` `^0.60.0`
**Status:** Resolved ✅ on branch `mike/feat/permissions` (see "Resolution" below)

## Summary

The local chat agent executed tools (including `write`, `edit`, `bash`) with no user approval. A permission system existed in the codebase — a dialog, an IPC channel, default ask/allow/deny rules, and a `SessionManager.requestPermission()` method — but nothing in the local-agent execution path ever invoked it. The only gate that actually fired locally was the sudo-password prompt for `bash` commands containing `sudo`.

## Impact (pre-fix)

- User expectation (set by `PermissionDialog.tsx` and `settings.permissionRules` defaults of `ask` for `write`/`edit`/`bash`) was that destructive tools prompt for approval.
- Reality: the model could write/edit files and run shell commands in the workspace (and inside the Lima/WSL2 sandbox if configured) without any confirmation.
- The README's disclaimer about "authorizing file modifications or deletions" implied a consent step that didn't exist for local sessions.

### 1. Permission plumbing existed but was never wired into local tool calls

- `SessionManager.requestPermission()` was defined in `src/main/session/session-manager.ts:1133` and sent `permission.request` IPC events to the renderer, where `PermissionDialog.tsx` handled them.
- Searching the entire `src/**/*.ts` for `requestPermission(` returned **only the definition** — zero callers.
- Searching for `permissionRules` returned only the type declaration (`src/renderer/types/index.ts:406`) and the defaults block (`src/renderer/store/index.ts:207`). Nothing in the main process read these rules.

### 2. `createAgentSession(...)` was called without any permission hook

In `src/main/claude/agent-runner.ts` (around L1899, pre-fix):

```ts
const { session: newPiSession } = await createAgentSession({
  model: piModel,
  thinkingLevel,
  authStorage,
  modelRegistry,
  tools: wrappedTools as unknown as ReturnType<typeof createCodingTools>,
  customTools: mcpCustomTools,
  sessionManager: PiSessionManager.inMemory(),
  settingsManager: PiSettingsManager.inMemory({ ... }),
  resourceLoader,
  cwd: effectiveCwd,
});
```

No `tool_call` handler, no per-tool `execute` wrapper that prompted. The only custom wrapping was `wrapBashToolForSudo`, which handled the sudo password case, not the general tool-approval case.

### 3. Remote path had approval — but only for Feishu/Slack

`src/main/remote/remote-manager.ts:618` consulted `config.gateway.autoApproveSafeTools`, auto-approved a read-only safelist, and otherwise posted a permission-request message back to the channel. The concept was implemented end-to-end — just not for local chat.

### Steps to Reproduce

1. Open the app and configure any model.
2. Leave `settings.permissionRules` at the defaults (which declared `write`/`edit`/`bash` as `ask`).
3. Ask the agent to modify a file in the workspace, e.g. "Add a TODO comment to `readme.md`".
4. **Observed:** the edit went through silently.
5. **Expected:** a `PermissionDialog` would appear first.

### Expected Behaviour

The local chat agent should consult `settings.permissionRules` before running any tool. Specifically, with the default rules shipped in `src/renderer/store/index.ts` (which declare `write` / `edit` / `bash` as `ask`):

1. When the agent attempts to run one of those tools, `PermissionDialog.tsx` should appear.
2. The user should be able to choose **Allow**, **Deny**, or **Always Allow**.
3. Rules set to `allow` (e.g. `read`, `glob`, `grep`, `ls`) should execute silently.
4. Rules set to `deny` should fail fast without prompting, returning a tool-result error the model can see.
5. Pattern-matching on `PermissionRule.pattern` should be honoured (e.g. `bash` + `ls *` = allow, other `bash` commands = ask).
6. `Always Allow` should apply for the rest of the session and be forgotten on session delete.

This is what the README's disclaimer about "authorizing file modifications or deletions" implies, and what remote (Feishu/Slack) sessions already do via `remote-manager.ts`.

### Actual Behaviour

The local chat agent executes every tool — including `write`, `edit`, and `bash` — with no user approval whatsoever. The permission system exists end-to-end in the codebase but is completely disconnected from the local tool-execution path:

- `PermissionDialog.tsx` exists and is mounted in the renderer, but never receives a `permission.request` event during a local session.
- `settings.permissionRules` exists as a type and has defaults, but nothing in the main process reads it.
- `SessionManager.requestPermission()` is defined in `src/main/session/session-manager.ts` and sends `permission.request` IPC events, but has **zero callers** in `src/**/*.ts` besides the definition itself.
- `createAgentSession(...)` in `src/main/claude/agent-runner.ts` is called without any `tool_call` hook, permission wrapper, or approval gate. The only tool-level wrapping is `wrapBashToolForSudo`, which handles the narrow sudo-password case, not general tool approval.

Net effect: the model can silently write/edit files anywhere in the workspace and run shell commands (including inside the Lima/WSL2 sandbox if configured) without any confirmation. The remote path (Feishu/Slack) does have approval via `config.gateway.autoApproveSafeTools` in `remote-manager.ts`, so the behaviour is inconsistent across session types.

```text
No errors are raised — this is a silent behaviour gap rather than a crash. The evidence is structural: searching the main process for callers of `requestPermission(` or readers of `permissionRules` returns empty.


# Search for callers of the permission method — finds only the definition
$ rg --type ts 'requestPermission\(' src/
src/main/session/session-manager.ts:1143:  async requestPermission(  ← definition only, 0 callers

# Search for readers of permission rules in main — finds nothing
$ rg --type ts 'permissionRules' src/main/
(no results)

# Search for readers of permission rules in the renderer — type + defaults only
$ rg --type ts 'permissionRules' src/renderer/
src/renderer/types/index.ts:406:  permissionRules?: PermissionRule[];
src/renderer/store/index.ts:207:    { tool: 'write', action: 'ask' },
src/renderer/store/index.ts:208:    { tool: 'edit', action: 'ask' },
src/renderer/store/index.ts:209:    { tool: 'bash', action: 'ask' },
...

# createAgentSession is called without any permission hook
# src/main/claude/agent-runner.ts (around L1899, pre-fix):
const { session: newPiSession } = await createAgentSession({
  model: piModel,
  thinkingLevel,
  authStorage,
  modelRegistry,
  tools: wrappedTools as unknown as ReturnType<typeof createCodingTools>,
  customTools: mcpCustomTools,  // ← no permission wrapper
  sessionManager: PiSessionManager.inMemory(),
  settingsManager: PiSettingsManager.inMemory({ ... }),
  resourceLoader,
  cwd: effectiveCwd,
});
# No tool_call handler, no execute() wrapper that prompts, no beforeToolCall hook.


For contrast, the remote path in `src/main/remote/remote-manager.ts:618` does consult `config.gateway.autoApproveSafeTools`, auto-approves a read-only safelist, and otherwise posts a permission-request message back to the channel — so the concept is implemented end-to-end for remote sessions, just not local.
```

## Proposed Fix

Two viable approaches — recommend the first:

**Option A — Install a permission hook via `agent.setBeforeToolCall(...)`** on the pi-coding-agent session immediately after `createAgentSession(...)`. This single interception point fires for both built-in tools (`read`, `bash`, `edit`, `write`) and MCP custom tools (the SDK ignores wrapped `execute` functions on built-in tools, so execute-wrapping alone doesn't cover them). Chain to the SDK's original hook on `allow`, return `{ block: true, reason }` on `deny`, and `await requestPermission(...)` on `ask`. Introduce a `src/main/config/permission-rules-store.ts` main-process cache, hydrate it from the renderer on startup + every `settings.update`, and clear session-scoped "always allow" memory on session delete.

**Option B — pi-coding-agent extension with a `tool_call` handler.** Cleaner long-term but requires writing a full extension scaffold because `ExtensionAPI.on(...)` is not reachable from `createAgentSession` options. More churn risk against the SDK's 0.x extension API.

obodnikov:
Fixes #164

## What this does

Wires the existing PermissionDialog, permission.request IPC channel, and settings.permissionRules into the local pi-coding-agent tool-execution path. Local sessions now prompt for approval on write/edit/bash and any MCP tool whose rule is `ask`, matching the behaviour remote (Feishu/Slack) sessions already had.

## Commits

1. **feat(permissions)** — Install a permission-gating hook on the pi-coding-agent session via `agent.setBeforeToolCall(...)`, chaining to the SDK's original hook. Covers both built-in tools (read, bash, edit, write) and MCP customTools in one interception point. Adds a main-process `permission-rules-store` with IPC-safe sanitization, hydrates it from the renderer on startup and every `settings.update`, and clears session-scoped "always allow" memory on session delete. Fails closed if the permission IPC rejects.

2. **fix(renderer)** — Follow-up fix for a React/preload integration bug that was masking the feature end-to-end. The preload's `window.electronAPI.on()` is a single-slot bridge; seven components call `useIPC()`, and each `useEffect` cleanup tore down the shared listener. When `PermissionDialog` unmounted after the user clicked Allow, the listener was removed and every subsequent `webContents.send()` call (tool result, session.status=idle) was silently dropped. Adds a module-level `ipcListenerInstalled` flag so only the first `useIPC()` caller installs the listener and registers a cleanup. Strict-mode-safe.

## Testing

Reproduced end-to-end on macOS with `deepseek/deepseek-v3.2:nitro` via OpenRouter:

1. Fresh session, prompt "Run `ls -la` in current directory"
2. `PermissionDialog` appears with the proposed bash command ✅
3. Click Allow ✅
4. Tool result appears in chat ✅
5. Agent's follow-up reply streams in ✅
6. Loading spinner clears when `session.status=idle` arrives ✅

Also verified Deny (tool returns isError) and Always Allow (subsequent same-tool calls skip the dialog in the same session).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -83,6 +83,8 @@ progress.md
 task_plan.md
 findings.md
 mcp_server.log
+adaptations/
+.index/
 
 # Development log
 DEV_LOG.md
--- a/src/main/claude/agent-runner.ts
+++ b/src/main/claude/agent-runner.ts
@@ -24,6 +24,7 @@ import { Type, type TSchema } from '@sinclair/typebox';
 import { getSharedAuthStorage, ModelRegistry } from './shared-auth';
 import type { Session, Message, TraceStep, ServerEvent, ContentBlock } from '../../renderer/types';
 import { v4 as uuidv4 } from 'uuid';
+import { decidePermission, rememberAlwaysAllow } from '../config/permission-rules-store';
 import { PathResolver } from '../sandbox/path-resolver';
 import { MCPManager } from '../mcp/mcp-manager';
 import { mcpConfigStore } from '../mcp/mcp-config-store';
@@ -401,6 +402,12 @@ interface AgentRunnerOptions {
     toolUseId: string,
     command: string
   ) => Promise<string | null>;
+  requestPermission?: (
+    sessionId: string,
+    toolUseId: string,
+    toolName: string,
+    input: Record<string, unknown>
+  ) => Promise<'allow' | 'deny' | 'allow_always'>;
 }
 
 interface CachedPiSession {
@@ -427,6 +434,12 @@ export class ClaudeAgentRunner {
     toolUseId: string,
     command: string
   ) => Promise<string | null>;
+  private requestPermission?: (
+    sessionId: string,
+    toolUseId: string,
+    toolName: string,
+    input: Record<string, unknown>
+  ) => Promise<'allow' | 'deny' | 'allow_always'>;
   private pathResolver: PathResolver;
   private mcpManager?: MCPManager;
   // @ts-expect-error stored for future plugin support
@@ -707,6 +720,7 @@ ${hints.join('\n')}
     this.sendToRenderer = options.sendToRenderer;
     this.saveMessage = options.saveMessage;
     this.requestSudoPassword = options.requestSudoPassword;
+    this.requestPermission = options.requestPermission;
     this.pathResolver = pathResolver;
     this.mcpManager = mcpManager;
     this._pluginRuntimeService = pluginRuntimeService;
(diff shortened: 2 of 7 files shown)


hqhq1025:
Maintainer rebase + tests pushed (CC @obodnikov, ready for final review).

## What changed in this push

**1. Rebase onto current `main`** — picked up #176, #177, #186, #196, #202 merged today.
- `session-manager.ts`: trivial 2-spot conflict in imports + `deleteSession()` cleanup. Resolved by keeping both: the new `extensionManager.onSessionDeleted()` call from #202 and `forgetSessionPermissions()` from this PR sit side by side. Also added `forgetSessionPermissions()` inside `batchDeleteSessions()` (legacy naming cleanup added a hooked path there that this PR's original diff didn't cover).
- `agent-runner.ts`: auto-merged cleanly. `installPermissionHook()`, `requestPermission` plumbing, and `decidePermission`/`rememberAlwaysAllow` imports all preserved over the legacy-naming refactor (#202) and `originalName` MCP changes (#176).
- `index.ts`, `useIPC.ts`, `permission-rules-store.ts`: no conflicts.

**2. Display-name integration (1 small follow-up)**
PR #176 added `originalName` to MCP tools — the human-readable name before OpenAI-compat sanitization. The permission hook now translates the canonical sanitized name through `getToolDisplayName()` before showing it to the user, so the dialog reads `chrome_screenshot` instead of `mcp__chrome__chrome_screenshot__ab12`. Rule matching and `rememberAlwaysAllow` still use the canonical name so allow-once decisions stay stable across calls.

**3. Unit tests — `src/tests/config/permission-rules-store.test.ts` (29 tests)**
Covers the security-critical surface:
- allow/deny/ask decisions (default + explicit rules)
- glob pattern matching (`ls *`, `*sudo*`, regex-metachar escaping)
- session-scoped always-allow: works in-session, doesn't leak across sessions, clears on `forgetSessionPermissions()`, case-insensitive
- fail-safe sanitation: `null`/non-array/empty-after-sanitize/garbage entries all fall back to `DEFAULT_RULES` (never an empty allow-everything ruleset); invalid `action` values coerce to `'ask'`
- `getPermissionRules()` returns shallow copies so callers can't mutate the cache

## Security review notes (hands-on, all green)

- `agent.setBeforeToolCall` fires for built-in tools (read/write/edit/bash) AND MCP customTools at a single interception point — verified via SDK source inspection
- Original SDK `_beforeToolCall` is captured before overwrite and chained on the allow path → tool_call extension events still flow
- IPC failure path inside the `try { await requestPermission(...) }` returns `{ block: true, reason }` — fails closed
- Always-allow lives in an in-process `Map<sessionId, Set<toolName>>` and is wiped on `deleteSession` / `batchDeleteSessions` via `forgetSessionPermissions()` — no cross-session leak, no disk persistence
- `setPermissionRules()` re-sanitizes every IPC payload from the renderer; `decidePermission()` defence-in-depth re-validates the matched rule's action

## Local checks

- `npm run lint` → 0 errors (8 pre-existing warnings unchanged)
- `npx tsc --noEmit` → 0 new errors (only pre-existing slack-channel errors)
- `npm run test` → 840 passing including 29 new (18 pre-existing failures in `memory/*` and `deepseek-thinking-serialization` are baseline, unrelated)

Branch force-pushed via `--force-with-lease`; no rewrites to your authorship.

PR #175, opened 2026-04-29
feat(loop-guard): add tool-call loop detection and protection

joekingss:
# Pull Request

## Problem
Models like gemini-3.1-pro with thinking disabled have been observed producing hundreds of repetitive tool calls in a single turn — sometimes 1500+ invocations of the same MCP tool, burning 370k+ tokens and running for 3-4 hours before the user notices or the API times out. There was no server-side safety net to detect or stop these loops.

### New module `src/main/claude/agent-runner-loop-guard.ts` (381 lines)

A **pure**, zero-dependency module (no Electron, no logger, no network) implementing two independent detection layers:

**Layer 1 — Hash-based group detection**
- Hash the entire tool-call list of each `message_end` using stable keys (MD5 over normalized JSON).
- Track hashes in a sliding window (default 20 entries).
- Stable key rules:
  - `read_file` → `${name}:${path}#bucket=${floor(start_line/200)}` (adjacent ±200 lines collapse)
  - `write_file` / `str_replace` / `edit_file` / etc. → `${name}:${md5(full input)}` (any content change = distinct)
  - Generic tools → extract key fields (`path`/`file_path`/`url`/`query`/`command`/`regex`/`pattern`/`keyword(s)`), md5 this subset; if none exist, md5 full input (ignores noise like `session_id`, `ts`).

**Layer 2 — Per-tool frequency detection**
- Tracks cumulative invocations of each `event.toolName` within the turn (no parameter comparison).
- Catches cross-parameter loops like repeatedly reading *different* files.

**Three-tier escalation thresholds** (all hardcoded defaults, configurable via constructor):

| Tier   | Hash repeats | Tool invocations |
|--------|-------------|------------------|
| warn   | 3           | 30               |
| halt   | 5           | 50               |
| abort  | 8           | 80               |

**Actions per tier:**
- **warn/halt** → fire-and-forget `piSession.sendUserMessage(steerText, { deliverAs: 'steer' })` injecting a user message into the stream. Warn is a soft suggestion; halt is a hard "stop all tools, output text now" instruction.
- **abort** → send an assistant text message with explanation + 3 suggestions (enable thinking / switch model / split task), emit `sendTraceUpdate(status: 'error')`, then `controller.abort()` to kill the session.
- Each tier fires **once per hash/tool name** (idempotent via `Set` de-dup). Abort marker is never reset; warn/halt markers reset when the hash slides out of the window, allowing fresh cycles.

### `agent-runner.ts` integration (+87/-3)

- Instantiate `LoopGuard` per `prompt()` call.
- In `message_end`: extract `toolCall` blocks from `resolvedPayload.effectiveContent`, call `loopGuard.recordAssistantMessage(descriptors)`.
- In `tool_execution_start`: call `loopGuard.recordToolInvocation(event.toolName)`.
- `handleLoopGuardDecision` callback: log warn, send steer/abort messages, abort controller. Guarded by `controller.signal.aborted` to prevent post-abort side effects.
- Runtime-compatible: checks `piSession.sendUserMessage` existence before calling (defensive against SDK version drift).

### Tests `tests/agent-runner-loop-guard.test.ts` (235 lines, 21 cases)

Full coverage of:
- `stableToolKey` — line bucketing, write-sensitivity, key-field extraction, key-order stability, null-input handling.
- `messageCallsHash` — deterministic hashing, order sensitivity, empty-list edge case.
- Layer 1 — warn/halt/abort at correct counts, cross-group isolation, empty-message ignore, sliding-window decay, adjacent read-file bucket collapse.
- Layer 2 — freq warn/halt/abort thresholds, per-tool independence, custom constructor overrides.
- Message builders — count/toolName interpolation, content presence.
- Default config constants match documented values.

## Cost impact
Based on a real incident (3h45min, 1597 tool calls, 370k tokens, ~$11 at $30/Mtok):

| Metric              | Before           | After (worst case) | Savings     |
|---------------------|------------------|--------------------|-------------|
| Max tool calls      | unbounded        | 80                 | ~95%        |
| Max tokens          | 370k+            | ~2-3k              | ~99%        |
| Max wall time       | 3-4 hours        | ~1-2 minutes       | ~99%        |

## Files changed
- `src/main/claude/agent-runner-loop-guard.ts` (new, 382 lines)
- `src/main/claude/agent-runner.ts` (+87/-0)
- `tests/agent-runner-loop-guard.test.ts` (new, 235 lines)

## Test plan
- [✓] `tsc --noEmit` passes
- [✓] `vitest run tests/agent-runner-loop-guard.test.ts` — 21/21 pass
- [✓] Hash layer: 3rd repeat → warn, 5th → halt, 8th → abort, then idempotent
- [✓] Freq layer: 30th → warn, 50th → halt, 80th → abort
- [✓] Sliding window: after 20 different messages, old hash resets and can re-trigger
- [✓] Empty tool-call list (pure text response) → no-op
- [✓] Different tools counted independently (read_file × 29 + bash × 29 → both warn on 30th)
- [✓] `piSession.sendUserMessage` runtime check prevents crash when SDK lacks the method

### DIFF
--- a/src/main/claude/agent-runner-loop-guard.ts
+++ b/src/main/claude/agent-runner-loop-guard.ts
@@ -1,0 +1,382 @@
+/**
+ * Loop guard — detects runaway tool-call loops inside a single agent turn.
+ *
+ * Two-layer strategy:
+ *
+ *   Layer 1  Hash-based group detection
+ *     - Hash the entire tool-call list of each assistant message (MD5 over stable keys).
+ *     - Track hashes in a sliding window (default 20 entries).
+ *     - 3 repeats  → warn  (inject a "please stop" steering message)
+ *     - 5 repeats  → halt  (inject a hard "stop now, produce text" steering message)
+ *     - 8 repeats  → abort (upstream gave up listening — kill the turn)
+ *
+ *   Layer 2  Per-tool frequency detection (no parameter comparison)
+ *     - Tracks cumulative invocations of each tool type within the turn.
+ *     - Catches cross-parameter loops such as repeatedly reading *different* files.
+ *     - 30 invocations → warn
+ *     - 50 invocations → halt
+ *     - 80 invocations → abort
+ *
+ * Stable tool key generation (for hashing):
+ *   - read_file            → `${tool}:${path}#bucket=${floor(startLine / 200)}`
+ *                            (adjacent line ranges collapse into the same key)
+ *   - write_file / str_replace / edit_file / create_file / search_replace
+ *                          → `${tool}:${md5(fullArgs)}`
+ *                            (any content change = distinct operation)
+ *   - everything else      → `${tool}:${md5(extracted key fields)}`
+ *                            where key fields = path/file_path/url/query/command/regex/pattern
+ *
+ * This module is **pure** — no electron, no logger, no network — so it is
+ * trivially unit-testable. The host (agent-runner.ts) is responsible for
+ * translating decisions into side effects (sendUserMessage steer, abort, etc.).
+ */
+
+import { createHash } from 'node:crypto';
+
+/** Configuration knobs. All fields are required once normalised. */
+export interface LoopGuardConfig {
+  /** How many most-recent assistant-message hashes to retain. */
+  messageHashWindow: number;
+  /** Same hash appearing this many times → soft warning. */
+  duplicateHashWarnThreshold: number;
+  /** Same hash appearing this many times → hard halt steering. */
+  duplicateHashHaltThreshold: number;
+  /** Same hash appearing this many times → unilateral abort. */
+  duplicateHashAbortThreshold: number;
+  /** Same tool invoked this many times in the turn → soft warning. */
+  toolFrequencyWarnThreshold: number;
+  /** Same tool invoked this many times in the turn → hard halt steering. */
+  toolFrequencyHaltThreshold: number;
+  /** Same tool invoked this many times in the turn → unilateral abort. */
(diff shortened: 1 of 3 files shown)


hqhq1025:
Hi @joekingss — thanks for this important feature. We want to ship it in v3.3.1.

Because `maintainerCanModify` is disabled on this PR (head branch is `Golden-Bridge-404:main`, your fork's default), we cannot push the two follow-up fixes directly to your branch. To avoid blocking the release, I've opened #225 which contains your original commit cherry-picked verbatim (commit author/date preserved, plus a `Co-authored-by` trailer) followed by two maintainer fix commits addressing the review findings:

1. `fix(loop-guard): use streak counter instead of frequency map` — fixes the false positive where `A/B/A/B/A` would push A's count to 3 and trip the warn threshold. Replaces the per-hash frequency map with a single (currentHash, currentStreak) pair that resets on any hash change. Adds regression test.
2. `fix(loop-guard): preserve error trace status across loop-guard abort` — fixes the case where the catch block was overwriting your guard's `status:'error'` trace with `status:'completed', title:'Cancelled'`. Adds an `abortedByLoopGuard` flag in the `prompt()` scope and a parallel branch in the AbortError catch. New test pins the wiring.

All credit for the design and the bulk of the implementation remains with you — both via the preserved cherry-pick authorship and the explicit Co-authored-by trailer on the rescue PR. #225 has `Closes #175` so merging it will close this one automatically.

Please take a look at #225 if you'd like to review the maintainer fixes; happy to iterate before merge.

PR #200, opened 2026-05-10
fix(mcp): sanitize MCP tool names for strict providers like DeepSeek

### LINKED ISSUE #159: DeepSeek API + MCP: 400 error due to invalid tool names (periods/colons not sanitized)
### App Version

Version 3.3.0 (3.3.0)

### Platform

macOS

### Bug Description

When using DeepSeek API as the model provider, enabling any MCP connector
(Firefox MCP, Xcode MCP, etc.) causes all requests to fail with a 400 error.

The error occurs because MCP servers expose tools with names containing
periods and colons (e.g., "browser.context", "devtools:inspect"). DeepSeek's
API strictly requires tool names to match the pattern ^[a-zA-Z0-9_-]+$.

Open CoWork passes the raw MCP tool names to DeepSeek without sanitization,
triggering the rejection before the request reaches the model.

This makes DeepSeek effectively unusable with MCP features.

### Steps to Reproduce

1. Open Open CoWork
2. Go to Settings → MCP Connectors
3. Add and enable Firefox MCP
4. Set the model provider to DeepSeek API
5. Ask: "List my open Firefox tabs"
6. Observe the 400 error

### Expected Behaviour

The Firefox MCP tool should execute successfully and return a list of open
browser tabs. DeepSeek should process the request normally with MCP tools.

### Actual Behaviour

Request fails immediately with HTTP 400 error:
"Invalid 'tools.function.name': string does not match pattern.
Expected a string that matches the pattern '^[a-zA-Z0-9_-]+$'."

No MCP tools can be used with DeepSeek enabled.

(the error is partly in Chinese)

```text
Error: 请求被上游拒绝（400），可能是模型/协议配置不兼容。
请检查模型名称、协议设置和 API 端点。
原始错误: 400 Invalid 'tools.function.name': string does not match pattern.
Expected a string that matches the pattern '^[a-zA-Z0-9_-]+$'.
```

### Additional Context

**_Open CoWork Modified Its Own Source Code to Fix This Bug_**

The issue documented below was identified and resolved through a directed session in which Open CoWork's AI agent was instructed to examine its own codebase to determine the cause of the DeepSeek MCP compatibility error.

The agent successfully extracted its own packaged app.asar source code from the running application. When standard extraction methods failed due to missing unpacked dependencies, the agent wrote custom Node.js scripts to bypass these failures and access the required files. It then traced the data flow from MCP tool registration through to the DeepSeek API request formatting, analyzing minified JavaScript to understand how tool names were being handled.

During this investigation, the agent identified and patched a secondary issue involving a strict mode compatibility flag. After determining that this was not the root cause, it continued searching and located the actual problem: raw MCP tool names containing periods and colons were being passed directly to the DeepSeek API without sanitization, violating the API's required ^[a-zA-Z0-9_-]+$ pattern.

The agent located the exact function—referred to internally as tTe—and the precise line where the raw tool name was being assigned. It applied a surgical one-line patch to sanitize the API-facing tool name by replacing invalid characters with underscores, while preserving the original display name in the UI label field. The agent then repacked the ASAR archive and replaced its own application bundle with the modified version.

After restarting Open CoWork, Firefox MCP tools executed successfully with DeepSeek as the model provider. The bug was resolved.

This occurred on a standard macOS installation using the publicly available Open CoWork application. The AI agent, under direction, read its own source code, identified a flaw in its own implementation, modified itself to correct that flaw, and continued operating with the fix in place.

The attached workspace folder contains the complete record: the extracted source code, the custom scripts written to access it, the applied patch, the repacked application bundle, and documentation of the process.

Technical details of the bug and the fix follow.
`

So the bug occurs when Open CoWork sends MCP tool definitions to the DeepSeek API. MCP servers (such as Firefox MCP) expose tools with names that contain periods and colons (e.g., browser.context, devtools:inspect). DeepSeek's API strictly validates tool names against the pattern ^[a-zA-Z0-9_-]+$ and rejects any name containing invalid characters with a 400 error.

Open CoWork passes the raw MCP tool name directly to the API without sanitization. The offending code is in the MCP tool registration function (referred to internally as tTe in the compiled source). The line in question is:

`name:n.name,label:n.name.replace(/^mcp__/,"").replace(/__/g," → ")`

Here, n.name is the raw tool name from the MCP server. This value is assigned directly to the name field that gets sent to the API.

The Fix

Sanitize the name field at the point of registration by replacing any character that is not a letter, number, underscore, or hyphen with an underscore:

`name:n.name.replace(/[^a-zA-Z0-9_-]/g,"_"),label:n.name.replace(/^mcp__/,"").replace(/__/g," → ")`

This changes a tool name like browser.context to browser_context for the API request, while preserving the original display name in the label field for the UI.

Why This Works

The name field is what DeepSeek validates. Sanitizing it ensures compliance with the required pattern.

The label field is used only for display in the Open CoWork interface. Leaving it unchanged maintains readable tool names for the user.

Applying the fix at the registration point means the sanitized name flows through the entire system, eliminating the need for bidirectional mapping when tool calls are returned.

Verification

After applying this patch, Firefox MCP tools execute successfully with DeepSeek as the model provider. The 400 error no longer occurs.

WORKSPACE DOWNLOAD

The complete workspace folder is available via Mega:

https://mega.nz/folder/SpQyCDJA#t6NcupAvBKusjrc_6TLUwg

open-cowork-self-modification-workspace.zip (700 MB)

Contents: extracted source code, custom extraction scripts, analysis scripts, the applied patch, original app.asar backup, repacked app.asar, and full documentation.

willytop8:
## Summary

Fixes #159

Tool names containing periods or colons (e.g. `my.tool`, `ns:action`) cause 400 errors from providers like DeepSeek that require names matching `^[a-zA-Z0-9_-]+$`. The API-facing prefixed name was built as `mcp__ServerName__<raw tool.name>`, passing illegal characters through to the model API.

**Changes:**
- Sanitize `tool.name` at the point of prefixed-name construction: `tool.name.replace(/[^a-zA-Z0-9_-]/g, '_')`
- Add `originalToolName` field to `MCPTool` to preserve the raw name for the actual MCP server call (the old prefix-stripping approach would have produced the sanitized name instead of the original, silently breaking tool execution)
- `callTool` now uses `tool.originalToolName` directly rather than re-deriving from the key

## Type of change

- [x] Bug fix (`fix`)

## Checklist

- [ ] Tests added or updated for the changed behaviour

## Testing

1. Configure an MCP server that exposes tools with dots or colons in their names (e.g. a server returning `"my.resource.read"`)
2. With the fix: the tool is registered as `mcp__ServerName__my_resource_read` (safe for DeepSeek/strict providers), and calling it correctly forwards `my.resource.read` to the MCP server
3. Without the fix: the API call fails with 400 from strict providers; or, with naive sanitization only, the MCP server call fails because it receives `my_resource_read` instead of `my.resource.read`

`tsc --noEmit` passes with zero errors.

### DIFF
--- a/src/main/mcp/mcp-manager.ts
+++ b/src/main/mcp/mcp-manager.ts
@@ -42,6 +42,7 @@ export interface MCPServerConfig {
  */
 export interface MCPTool {
   name: string;
+  originalToolName: string; // raw name from the MCP server — used when calling the server
   description: string;
   inputSchema: {
     type: string;
@@ -1355,10 +1356,13 @@ export class MCPManager {
           // Format: mcp__<ServerName>__<toolName> (double underscores, preserve case)
           // Sanitize: collapse whitespace and collapse any accidental __ sequences
           const serverKey = config.name.replace(/\s+/g, '_').replace(/__/g, '_');
-          const prefixedName = `mcp__${serverKey}__${tool.name}`;
+          // Sanitize tool.name for strict providers (e.g. DeepSeek) that require ^[a-zA-Z0-9_-]+$
+          const sanitizedToolName = tool.name.replace(/[^a-zA-Z0-9_-]/g, '_');
+          const prefixedName = `mcp__${serverKey}__${sanitizedToolName}`;
 
           newTools.set(prefixedName, {
             name: prefixedName,
+            originalToolName: tool.name, // preserved for the actual MCP server call
             description: tool.description || '',
             inputSchema: {
               type: 'object',
@@ -1426,15 +1430,8 @@ export class MCPManager {
       throw new Error(`MCP tool not found: ${toolName}`);
     }
 
-    // 提取实际工具名（格式：mcp__<ServerName>__<toolName>）
-    let actualToolName = toolName;
-    if (toolName.startsWith('mcp__')) {
-      const remainder = toolName.slice('mcp__'.length);
-      const separatorIndex = remainder.indexOf('__');
-      if (separatorIndex !== -1) {
-        actualToolName = remainder.slice(separatorIndex + 2);
-      }
-    }
+    // Use the stored original name so dots/colons in tool names survive the sanitization pass
+    const actualToolName = tool.originalToolName;
 
     logCtx(`[MCPManager] Calling tool ${actualToolName} on server ${tool.serverName}`);
 
--- a/src/main/mcp/mcp-manager.ts
+++ b/src/main/mcp/mcp-manager.ts
@@ -1357,8 +1357,19 @@ export class MCPManager {
           // Sanitize: collapse whitespace and collapse any accidental __ sequences
           const serverKey = config.name.replace(/\s+/g, '_').replace(/__/g, '_');
           // Sanitize tool.name for strict providers (e.g. DeepSeek) that require ^[a-zA-Z0-9_-]+$
-          const sanitizedToolName = tool.name.replace(/[^a-zA-Z0-9_-]/g, '_');
-          const prefixedName = `mcp__${serverKey}__${sanitizedToolName}`;
+          // Collapse consecutive underscores so spaces don't erode the __ separator convention.
+          const sanitizedToolName =
+            tool.name
+              .replace(/[^a-zA-Z0-9_-]/g, '_')
+              .replace(/_+/g, '_')
+              .replace(/^_|_$/g, '') || '_unnamed_';
+          // Disambiguate collisions: two names that sanitize identically get a numeric suffix.
+          let finalName = sanitizedToolName;
+          let counter = 1;
+          while (newTools.has(`mcp__${serverKey}__${finalName}`)) {
+            finalName = `${sanitizedToolName}_${counter++}`;
+          }
+          const prefixedName = `mcp__${serverKey}__${finalName}`;
 
           newTools.set(prefixedName, {
             name: prefixedName,
--- a/tests/mcp-tool-name.test.ts
+++ b/tests/mcp-tool-name.test.ts
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
 import { describe, it, expect, vi } from 'vitest';
 
 vi.mock('electron', () => ({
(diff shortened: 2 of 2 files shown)


willytop8:
All three items from the bot review have been addressed in the latest commit (`70f543e`):

**[Major] Name collision after sanitization**
Added a disambiguation loop that appends a numeric suffix (`_1`, `_2`, …) until the prefixed key is unique in `newTools`. Two tools that sanitize to the same name (e.g. `my.tool` and `my-tool` → both `my_tool`) now register as `mcp__Server__my_tool` and `mcp__Server__my_tool_1` without silent overwrites.

**[Minor] Double underscores from spaces**
Chained `.replace(/_+/g, '_')` and `.replace(/^_|_$/g, '')` after the character-class replacement. Spaces in a tool name now collapse to a single underscore rather than bleeding into the `__` separator.

**[Nit] Empty sanitized name**
Added a `|| '_unnamed_'` fallback so a tool name made entirely of special characters (e.g. `..`) produces a valid, identifiable key instead of a broken `mcp__Server__`.

**Tests**
Added three new cases to `tests/mcp-tool-name.test.ts`:
- Dots/colons sanitized in the key, original name preserved for the actual MCP call
- Collision: `my.tool` and `my-tool` register as distinct keys and route correctly
- Empty-name fallback: `..` maps to `_unnamed_` and still routes to the correct original

hqhq1025:
Thanks for jumping on this, @willytop8! 🙏

After reviewing alongside #176 (which addresses the same issue), we've decided to merge #176 because it covers additional edge cases that surface in production:

- **OpenAI 64-character function-name limit** — sanitized names like `mcp__LongServerName__long_tool_name` can easily exceed this and get rejected by OpenAI / OpenAI-compatible providers
- **Deterministic deduplication** — sorts tools alphabetically before assigning `_2`, `_3` suffixes so session history stays stable across reconnects
- **Hash fallback** for absurdly long suffixes
- **UI display name plumbing** — keeps the human-readable `browser.context` in the UI while sending the sanitized `browser_context` to the model

Your fix was on the right track and your test additions were great — the core logic (sanitize + preserve `originalToolName`) matches what we're merging. We really appreciate you stepping in to address open issues. Hope you'll send more PRs! 🚀

Closing in favor of #176.

PR #201, opened 2026-05-10
fix(mcp): increase listTools timeout from 10s to 5 minutes

### LINKED ISSUE #161: MCP server listTools() timeout too short for complex operations (30s → 5min)
### App Version

3.3.0

### Platform

macOS

### Bug Description

The MCPManager class enforces a 30-second timeout on the listTools() operation (refreshTools). This is too short for complex MCP servers like Frida, which require more time to enumerate their tool sets. The timeout causes Frida MCP operations to fail with "listTools timeout after 30s" before any actual tool calls can be made.

This issue also affects other complex MCP servers, including Radare2 MCP, which times out during initial binary analysis and tool enumeration. Any MCP server that requires more than 30 seconds to initialize or enumerate its tools will be impacted by this hardcoded limit. The problem is not specific to Frida ... it is a systemic limitation in Open CoWork's MCP client implementation that affects all complex or slow-initializing MCP servers.

### Steps to Reproduce

1. Install Frida MCP server
2. Add it as an MCP connector in Open CoWork
3. Attempt to use any Frida MCP tool (e.g., enumerate_processes)
4. Observe the timeout error after 30 seconds

### Expected Behaviour

Frida MCP should successfully enumerate its tools and execute commands without artificial timeout restrictions, or at least have a configurable/reasonable timeout for complex operations.

### Actual Behaviour

Frida MCP operations fail with "listTools timeout after 30s". The 30-second limit is insufficient for Frida to complete its tool discovery.

```text
Tool call timeout after 30000ms
listTools timeout after 30s
```

### Additional Context

Root cause identified: In dist-electron/main/index-DnBwUUuJ.js, the MCPManager class defines:

- refreshTools: const r=3e4 (30,000ms = 30 seconds) for listTools()
- callTool: const p=3e5 (300,000ms = 5 minutes) for tool execution

The refreshTools timeout is the bottleneck. Frida MCP requires more than 30 seconds to enumerate its tools, especially on first connection.

Fix applied locally and verified working:
- Changed refreshTools: const r=3e4 → const r=3e5 (5 minutes)
- Changed callTool: const p=3e5 → const p=6e5 (10 minutes) for extra buffer

After patching app.asar with these changes, Frida MCP works without timeout errors. The changes are minimal, backwards-compatible, and significantly improve MCP reliability for complex servers.

Suggestion: Either increase the default refreshTools timeout to 300000ms (5 minutes), or make MCP timeouts configurable per-server via the connector settings.

Full investigation and patch details available if needed.

willytop8:
## Summary

Fixes #161

The `listTools()` call had a 10-second timeout. Complex MCP servers (those that spawn subprocesses, enumerate many resources, or download dependencies at startup) routinely exceed this limit, causing spurious tool-refresh failures on first connect.

**Changes:**
- `src/main/mcp/mcp-manager.ts`: raise `timeoutMs` from `10_000` to `300_000` (5 minutes)
- Update the rejection message to be dynamic (`listTools timeout after ${timeoutMs}ms`) rather than hard-coding "10s"

Note: the issue description mentioned 30s as the existing value; the actual code had 10s for `listTools` (30s was the separate `callTool` timeout). Only `listTools` is changed here.

## Type of change

- [x] Bug fix (`fix`)

## Checklist

## Testing

Connect an MCP server that takes more than 10 seconds to enumerate its tools (e.g. one that runs `npx` on first launch). Confirm tools load successfully instead of failing with a timeout error.

`tsc --noEmit` passes with zero errors.

### DIFF
--- a/src/main/mcp/mcp-manager.ts
+++ b/src/main/mcp/mcp-manager.ts
@@ -1330,11 +1330,14 @@ export class MCPManager {
         if (!config) continue;
 
         // Add timeout for listTools call to prevent hanging
-        const timeoutMs = 10000; // 10 second timeout
+        const timeoutMs = 300_000; // 5 minute timeout — complex MCP servers can take time to enumerate
         const listToolsPromise = client.listTools();
         let timeoutId: ReturnType<typeof setTimeout>;
         const timeoutPromise = new Promise<never>((_, reject) => {
-          timeoutId = setTimeout(() => reject(new Error('listTools timeout after 10s')), timeoutMs);
+          timeoutId = setTimeout(
+            () => reject(new Error(`listTools timeout after ${timeoutMs}ms`)),
+            timeoutMs
+          );
         });
 
         log(`[MCPManager] Fetching tools from ${config.name} (timeout: ${timeoutMs}ms)...`);


hqhq1025:
Thanks for the quick fix, @willytop8! 🙏

After reviewing alongside #177 (which addresses the same underlying issue), we've decided to merge #177 because it solves two additional problems that the simple timeout bump leaves on the table:

- **Parallel `listTools` discovery** — `refreshTools()` currently iterates servers serially, so a single 5-minute-hanging server blocks every other MCP server from refreshing. #177 switches to `Promise.allSettled`.
- **Shared deadline across `callTool` retries** — `callTool` retries timeouts up to 3 times, so a single hung tool would now make users wait up to **15 minutes** before seeing an error. #177 uses a shared deadline so total wait stays ≤ 5 minutes.
- **Regression coverage** with fake-timer tests for all three scenarios.

Your one-line bump correctly identified the root cause (10s is too tight) — we just need the broader fix to avoid making the UX worse on the failure path. Really appreciate you tackling open issues! Hope you'll send more PRs. 🚀

Closing in favor of #177.

PR #249, opened 2026-06-16
fix(sandbox): enrich PATH before Lima detection so packaged macOS app finds limactl

### LINKED ISSUE #248: Sandbox unavailable on macOS: PATH not enriched before Lima detection in packaged app
**App Version:** 3.3.1

**Platform:** macOS (reproduced on Apple Silicon; also affects Intel)

## Bug Description
On the packaged macOS app, the sandbox stays disabled with **"Lima not installed"** even when Lima is installed (`brew install lima`) and `limactl` works fine in a terminal. Settings → Sandbox reports the sandbox as unavailable.

## Steps to Reproduce
1. `brew install lima` — `limactl` lands in `/opt/homebrew/bin` (Apple Silicon) or `/usr/local/bin` (Intel).
2. Launch the packaged Open Cowork app from Finder/Dock (not from a terminal).
3. Open Settings → Sandbox.
4. Sandbox is reported unavailable / "Lima not installed", even though `which limactl` succeeds in Terminal.

## Expected Behaviour
Lima is detected and the sandbox is available, since `limactl` is installed and on the user's shell PATH.

## Actual Behaviour
Sandbox reported unavailable. Logs show:

```
[Lima] limactl not found - Lima not installed
```

## Root Cause
GUI apps launched from Finder/Dock on macOS inherit a stripped launchd PATH (`/usr/bin:/bin:/usr/sbin:/sbin`) that excludes `/opt/homebrew/bin` and `/usr/local/bin`. `LimaBridge.checkLimaStatus()` runs `execAsync('which limactl')` (`src/main/sandbox/lima-bridge.ts:142`), which relies on `process.env.PATH` and therefore fails.

The app already ships `enrichProcessPathForBuild()` (`src/main/claude/agent-runner.ts`) which restores the login-shell PATH — but it only runs lazily before the first `createCodingTools()` (`agent-runner.ts:2146`), i.e. when an agent task starts. The sandbox detection runs *earlier*: at startup via `startSandboxBootstrap()` → `bootstrapLima()` → `checkLimaStatus()` (`src/main/sandbox/sandbox-bootstrap.ts:305`), and again from the `sandbox.checkLima` IPC when the Sandbox settings panel opens. At those points PATH is still stripped, so detection fails.

## Suggested Fix
Run the existing (idempotent) PATH enrichment once at app startup, before the sandbox bootstrap, so the Homebrew bin dir is on PATH whenever Lima is probed. A PR implementing this follows.

## Workaround (until fixed)
Set the GUI PATH for the current login session and persist it:
```
launchctl setenv PATH "/opt/homebrew/bin:$PATH"
```
…plus a LaunchAgent to reapply at login, or simply launch the app from a terminal that has Homebrew on PATH.

craigamcw:
## Summary

Fixes #248.

On the packaged macOS app the sandbox is reported as unavailable ("Lima not installed") even when Lima is installed, because the Lima detection runs with the stripped launchd PATH that GUI apps inherit (no `/opt/homebrew/bin` / `/usr/local/bin`).

The app already has an idempotent `enrichProcessPathForBuild()` that restores the login-shell PATH, but it only runs lazily before the first `createCodingTools()` (an agent task). The sandbox detection (`checkLimaStatus()` → `which limactl`) runs earlier — at startup via `startSandboxBootstrap()` and from the `sandbox.checkLima` IPC — so PATH was still stripped at detection time.

This PR exports `enrichProcessPathForBuild()` and `await`s it once at the top of the `app.whenReady()` handler, before the sandbox bootstrap and any IPC can run. Because the function is guarded by `pathEnriched`, the existing call before `createCodingTools()` becomes a no-op. Dev mode and the Windows registry-based PATH restore are unaffected.

## Type of change

- [x] Bug fix (`fix`)

## Checklist

- [x] Code follows the project style (TypeScript strict, ESLint, Prettier) — changed lines only
- [ ] `npm run test` passes locally — see Testing (new test + typecheck pass; full suite not run locally)
- [ ] `npm run lint` passes locally — see Testing (no new errors from this change; pre-existing findings noted)
- [ ] UI changes tested on both macOS and Windows — N/A, no UI; macOS-specific startup wiring, Windows path untouched
- [ ] New user-facing strings added to i18n files — N/A, no user-facing strings

## Testing

Repro and root cause are detailed in #248. Branch is based on `dev`.

Verified locally:
- `npx tsc --noEmit` — the changed files type-check cleanly. (There is one unrelated pre-existing error on `dev`, `src/main/config/config-store.ts:369 'getConfigKey' is declared but never read`, untouched by this change.)
- New regression test `tests/startup-path-enrichment.test.ts` (4 assertions) — passes via `npx vitest run`. It follows the existing source-assertion style in `tests/agent-runner-pi.test.ts` and guards that the enrichment is exported, idempotent, imported by the entrypoint, and awaited within `whenReady()`.
- `eslint` on the two changed files — **no errors on the changed lines**. (These large files contain pre-existing `@typescript-eslint/no-explicit-any` findings unrelated to this change; left untouched to keep the diff minimal and focused.)

Note: the full `npm run test` suite was not run locally because dependencies were installed with `--ignore-scripts` (native `better-sqlite3` not rebuilt). CI will run the complete suite.

Manual verification on a real macOS install: before the change, Settings → Sandbox showed "Lima not installed" despite `which limactl` succeeding in Terminal (Homebrew `limactl`). Restoring the login-shell PATH at startup makes the detection resolve `limactl` correctly.

### DIFF
--- a/src/main/claude/agent-runner.ts
+++ b/src/main/claude/agent-runner.ts
@@ -145,11 +145,14 @@ function resolveBundledToolsBinDir(): string | null {
  * 3. Deduplicates all entries
  * 4. Writes the result back to `process.env.PATH`
  *
- * Called once before the first `createCodingTools()` — subsequent calls are no-ops.
+ * Idempotent: enriches once and caches the result, so it is safe to call from
+ * multiple entry points. Call at app startup (before the sandbox bootstrap /
+ * Lima detection) and before the first `createCodingTools()` — subsequent calls
+ * are no-ops.
  */
 let pathEnriched = false;
 
-async function enrichProcessPathForBuild(): Promise<void> {
+export async function enrichProcessPathForBuild(): Promise<void> {
   if (pathEnriched) return;
   pathEnriched = true;
 
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -38,6 +38,7 @@ import { SandboxSync } from './sandbox/sandbox-sync';
 import { WSLBridge } from './sandbox/wsl-bridge';
 import { LimaBridge } from './sandbox/lima-bridge';
 import { getSandboxBootstrap } from './sandbox/sandbox-bootstrap';
+import { enrichProcessPathForBuild } from './claude/agent-runner';
 import type { MCPServerConfig } from './mcp/mcp-manager';
 import type {
   ClientEvent,
@@ -759,6 +760,14 @@ function sendToRenderer(event: ServerEvent) {
 app
   .whenReady()
   .then(async () => {
+    // Enrich PATH before anything else so Homebrew-installed CLIs (notably
+    // `limactl`) are discoverable by the sandbox bootstrap and the
+    // `sandbox.checkLima` IPC. In the packaged macOS app, GUI processes inherit
+    // a stripped launchd PATH that excludes /opt/homebrew/bin and
+    // /usr/local/bin, which otherwise makes `which limactl` fail and reports
+    // the sandbox as unavailable even when Lima is installed.
+    await enrichProcessPathForBuild();
+
     // Apply dev logs setting from config
     const enableDevLogs = configStore.get('enableDevLogs');
     setDevLogsEnabled(enableDevLogs);
--- a/tests/startup-path-enrichment.test.ts
+++ b/tests/startup-path-enrichment.test.ts
@@ -1,0 +1,46 @@
+import { describe, expect, it } from 'vitest';
+import { readFileSync } from 'node:fs';
+import path from 'node:path';
+
+const read = (rel: string): string => readFileSync(path.resolve(process.cwd(), rel), 'utf8');
+
+/**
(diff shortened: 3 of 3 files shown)


craigamcw:
Heads-up for reviewers: the failing **`pr-review`** check here is **not** caused by this change.

The `Codex PR Review` workflow (`pull_request_target`) checks out the PR base ref (`dev`, via `ref: ${{ github.event.pull_request.base.ref }}`) and then runs `node .github/scripts/deepseek-pr-review.mjs` — but `.github/scripts/` exists only on `main`, not on `dev`. So the step fails with `MODULE_NOT_FOUND` for **any** PR targeting `dev`, independent of its contents. Filed #250 with the details and a couple of fix options.

This PR itself is just the 3 files in the diff; `eslint` (changed files), `tsc --noEmit`, and the added test pass locally. Happy to retarget to `main` if you'd prefer.

PR #253, opened 2026-06-16
feat(i18n): add 10 European language locales

craigamcw:
## What
Adds full UI translations for **10 European languages** alongside the existing English and Chinese:

`es` (Español), `fr` (Français), `de` (Deutsch), `it` (Italiano), `uk` (Українська), `pl` (Polski), `sv` (Svenska), `no` (Norsk), `nl` (Nederlands), `ro` (Română)

## Changes
- New locale files under `src/renderer/i18n/locales/` — each a complete translation of `en.json` (all 866 keys), with native orthography and every `{{interpolation}}` placeholder / pluralization key preserved.
- `src/renderer/i18n/config.ts` — import the locales and extend `supportedLngs`.
- `src/renderer/components/settings/SettingsGeneral.tsx` — add the languages to the language switcher (native names).

The existing `i18next-browser-languagedetector` picks these up automatically, and users can also select them in **Settings → General → Language**.

## Notes
- Verified: all 10 locales parse as valid JSON, have full key parity with `en.json`, and `tsc --noEmit` passes.
- Brand/technical tokens (Open Cowork, API, MCP, URL, model IDs, provider names, etc.) are intentionally left untranslated.
- Complements the Memory-settings English-fallback fix in #246 — the new locales include those keys, so they render correctly once #246 lands.


craigamcw:
Thanks for the thorough review — all points addressed in the latest commit:

**[MAJOR] Active-language highlight** — Fixed. `SettingsGeneral.tsx` no longer uses the `startsWith('zh') ? 'zh' : 'en'` logic; it now normalizes to the base code (`es-ES → es`, `zh-CN → zh`) and maps Norwegian `nb`/`nn` → the shipped `no` locale, so every language button highlights correctly.

**[MINOR] config.ts comment noise** — Reverted. The original Chinese comments are restored; only the functional changes (locale imports/resources, `supportedLngs`, and the new fallback options) remain in the diff.

**[NIT] Key-parity test** — Added `src/tests/i18n/locale-key-parity.test.ts`, which asserts every locale has *exactly* the same key set as `en.json` (fails on missing or extra keys). To make this hold, I also added the `memory.*` / `api.*` keys to `en.json`/`zh.json` so all 12 locales are consistent — these mirror #246, so this PR now effectively supersedes its data changes (happy to rebase/close #246 whichever you prefer).

**[Question] Region aliases / `nb`** — Added `fallbackLng: { nb: ['no'], nn: ['no'], default: ['en'] }` plus `nonExplicitSupportedLngs: true`, so `nb`/`nb-NO`/`nn` resolve to Norwegian and region variants of the other languages resolve to their base locale.

`tsc --noEmit` is clean and the new parity test passes (11/11).

SatoshiKomura:
Thanks for putting this together — really nice to see the **i18n** infrastructure extended like this.

  Would you consider adding **Japanese (ja)** as well? Japanese-speaking users currently rely on English, and a ja.json following the same approach (full key parity with en.json, brand/technical tokens left untranslated) would be a great fit alongside these European locales.

I'd be glad to help with translation review or contribute a ja.json myself if that's welcome.

PR #289, opened 2026-07-11
fix: remove inline <think> tag parser and add frontend escaping

### LINKED ISSUE #285: <think> stripping applied to every provider hides literal <think>…</think> content from non-reasoning models
### App Version

Reproduced against current source `a4f4c47` (present in 3.3.1).

### Platform

Both

### Bug Description

`resolveMessageEndPayload` runs every assistant text block through `splitThinkTagBlocks` (`src/main/agent/agent-runner-message-end.ts:183-199`) with no provider/model gate. That parser is scoped by its own docstring to "Ollama reasoning models (deepseek-r1, qwen3, etc.)" (`src/main/agent/think-tag-parser.ts:1-4`).

So when a provider that does not use inline `<think>` reasoning (Anthropic / OpenAI / Gemini) emits a literal `<think>…</think>` in its answer, the inner text is silently removed from the visible response and hidden in a collapsed thinking block. The same unconditional handling is on the live stream path (`src/main/agent/agent-runner.ts:2618`), so both the live view and the persisted message are affected identically.

This is silent data loss from the visible answer, not just a display quirk: for a code or markdown sample that contains the token, the code shown in the visible pane is incomplete.

### Steps to Reproduce

`splitThinkTagBlocks` is a pure function, so this is deterministic:

```ts
import { splitThinkTagBlocks } from './src/main/agent/think-tag-parser';

const input =
  'Ollama models hide reasoning like this: <think>step 1; step 2</think> and then give the answer.';
console.log(splitThinkTagBlocks(input));
```

### Expected Behaviour

One text block with the full sentence visible — the model did not use `<think>` for hidden reasoning, it is ordinary content.

### Actual Behaviour

Reproduced on source `a4f4c47`:

```
[
  { type: 'text',     text: 'Ollama models hide reasoning like this: ' },
  { type: 'thinking', thinking: 'step 1; step 2' },
  { type: 'text',     text: ' and then give the answer.' }
]
```

The visible text becomes `"Ollama models hide reasoning like this:  and then give the answer."` — `step 1; step 2` is gone from the answer.

### Root Cause

The split is correct for Ollama / deepseek-r1 (which stream reasoning inside `<think>`), but it is applied to every provider. Anthropic, OpenAI, and Gemini deliver reasoning via structured blocks or a separate `reasoning` field and never emit inline `<think>` as reasoning, so applying the strip to them can only corrupt legitimate content that happens to contain that token — most naturally when the user asks the model to explain or demonstrate the `<think>` format, or when a code sample includes it.

### Question / proposed fix

Is the global stripping intended (for example, to catch relays that inject `<think>` regardless of model), or should it be gated to reasoning providers? If gating is the right call, the fix is to run `splitThinkTagBlocks` only when the active model is a `<think>`-style reasoning model, leaving other providers' text verbatim. Happy to open a PR with a regression test once you confirm the preferred approach.

Reproduced against source `a4f4c47` in a clean container. Prepared with AI assistance.

max-cyq:
### Summary

Fixes #285.

Removes the inline `<think>...</think>` tag parser (`think-tag-parser.ts`)
and the unconditional `splitThinkTagBlocks` post-processing loop from
`agent-runner-message-end.ts`. Adds frontend escaping in `ThinkingBlock`
and `ContentBlockView` to prevent the markdown sanitizer from stripping
literal `<think>` references.

### Root Cause

Two separate issues:

1. **Backend parser was unnecessary.** The Ollama API (`/v1/chat/completions`)
   separates reasoning into a dedicated `reasoning` field mapped to
   `thinking_delta` by the pi-ai SDK. No provider emits inline `<think>` tags
   in the `content` field. The parser only caused harm by mis-parsing literal
   `<think>` text when models mention it in explanations.

   Evidence: Ollama's `thinking.Parser`
   ([source](https://github.com/ollama/ollama/blob/main/thinking/parser.go))
   strips `<think>...</think>` server-side before the response reaches any client.

2. **Frontend sanitizer stripped literal `<think>`.** `ReactMarkdown` with
   `rehypeSanitize` treats `<think>` as an unknown HTML tag and strips it.
   Converting to `&lt;think&gt;` prevents stripping while browsers correctly
   render `&lt;` as `<`.

### Changes

| File | Change |
|---|---|
| `src/main/agent/think-tag-parser.ts` | **Deleted** (-165 lines) |
| `src/main/agent/agent-runner.ts` | Remove thinkParser import & usage, simplify text_delta (-48 lines) |
| `src/main/agent/agent-runner-message-end.ts` | Remove unconditional splitThinkTagBlocks loop (-26 lines) |
| `src/renderer/components/message/ThinkingBlock.tsx` | Add escapeThinkTags() (+6 lines) |
| `src/renderer/components/message/ContentBlockView.tsx` | Apply escapeThinkTags to assistant text blocks (+0) |
| `tests/think-tag-parser.test.ts` | **Deleted** (-211 lines) |
| `tests/agent-runner-message-end.test.ts` | Add 3 tests for <think> preservation (+55 lines) |

**Net: +73 / -450 lines**

### Testing

```
npm run test  → 606/607 passed (1 pre-existing failure in path-guard.test.ts, unrelated)
tsc --noEmit → clean
npm run lint → 0 errors
```

### Checklist

- [x] Bug fix
- [x] Tests added
- [x] tsc --noEmit passes
- [x] lint passes
- [x] UI tested on macOS


Sun-sunshine06:
Addressed the automated review feedback in `d839841`:

- added focused coverage for opening/closing tags, malformed input, and already-escaped entities
- removed the stale flush parameters, delta fields, and unreachable emission branches left after deleting the stream parser

Validation:
- `npx vitest run tests/agent-runner-message-end.test.ts tests/escape-think-tags.test.ts --reporter=verbose` (34 passed)
- `npm run typecheck`
- `npm run lint` (0 errors; existing warnings only)

PR #282, opened 2026-07-06
feat: Agent Platform Phase 3 — UI/UX + StdioChannel + Config Write

hqhq1025:
## Summary

Completes all Phase 3 items from the Agent Platform roadmap (#274):

- **Headless Phase 2**: StdioChannel for bidirectional RPC (`--mode stdio`)
- **Config Phase 3**: `config_write` tool with permission-gating + sensitivity blocklist
- **Subagent Phase 3**: Real-time subagent progress rendering in chat stream
- **Compact Phase 3**: Context usage bar, compaction history panel, manual compact button

All implementations have been adversarially reviewed and critical/major bugs fixed before merge.

### Headless: StdioChannel (`src/main/remote/channels/stdio-channel.ts`)
- New `IChannel` impl reading/writing JSONL on stdin/stdout
- Protocol: `session.start` / `session.message` / `session.abort` → structured events
- Wire `--mode stdio` in headless path through RemoteManager
- EPIPE handling, graceful shutdown with session abort on stdin close
- 21 tests

### Config: `config_write` tool (`src/main/config/config-extension.ts`)
- Allow-list of safe writable fields (model, theme, contextWindow, etc.)
- Blocklist: apiKey, profiles, configSets, pattern-matched sensitive fields
- Permission: always-ask (never auto-approved, blocked in subagents)
- Fixes pre-existing subagent permission bypass (ask → allow silently)

### Subagent UI (`src/renderer/components/SubagentProgress.tsx`)
- Collapsible card showing lifecycle: spinner → tool activity → result
- Module-level singleton state with RAF-batched notifications
- Auto-cleanup after 5s, session-scoped tracking
- i18n (en/zh)

### Compact UI (`src/renderer/components/ContextUsageBar.tsx`)
- Real-time token progress bar with color transitions (green/yellow/red)
- "Compact Now" button with portal-based confirmation dialog
- Compaction history panel with expandable summaries
- Manual compact now emits `compaction.result` event (was silently succeeding)

## Adversarial Review Findings Fixed

| Component | Critical | Major | Fixed |
|-----------|----------|-------|-------|
| Subagent UI | 2 | 3 | All 5 |
| Compact UI | 2 | 5 | All 7 |
| StdioChannel | 2 | 4 | All 6 |
| Config Write | 0 | 0 | N/A (clean) |

Key fixes: dead-code event routing (SM bypassed eventSender), memo() stale refs, text_delta corruption, missing responseCallback, EPIPE crash, permission bypass.

## Test plan

[...]

PR #292, opened 2026-07-11
fix(remote): multi-turn session mapping (#291) + headless stdio stdin-close exit

### LINKED ISSUE #291: Multi-turn remote sessions fail: "No actual session ID found" after first turn (StdioChannel/RemoteManager)
## Summary

Multi-turn remote sessions break after the first turn. Any follow-up message on an existing remote session fails with:

```
[MessageRouter] Error processing message: Error: No actual session ID found for remote session: remote-<ts>-<n>
```

The client receives a generic `❌ An internal error occurred. Please try again later.` This affects **every remote channel** that routes through `RemoteManager` (the new `--headless --mode stdio` StdioChannel from #282, and the same code path used by the gateway channels). Single-shot (first) turns work; conversation continuity and `session.abort` on an existing session do not.

## Root cause

`RemoteManager` keeps two structures for the "is this a new session?" decision in `executeAgent()`:

- `remoteSessionIds: Set<string>` — remembers which remote session ids have been started. **Only ever added to, never removed.**
- `reverseSessionIdMapping: Map<remoteId, actualSessionId>` — needed to continue an existing session.

`clearSessionBuffer(actualSessionId)` is called on **every turn completion** (on `session.status` = `idle`/`error`, see `src/main/index.ts`). Besides flushing the ephemeral per-turn response buffers (its intended job), it also tears down the **persistent** session mappings:

```ts
// src/main/remote/remote-manager.ts  (clearSessionBuffer)
this.sessionIdMapping.delete(actualSessionId);
// ...deletes reverseSessionIdMapping / sessionChannelMapping / sessionOwnerMapping too
```

So after turn 1 completes:

- `reverseSessionIdMapping` / `sessionIdMapping` are emptied, **but**
- `remoteSessionIds` still contains the remote id.

On turn 2, `executeAgent()` computes `isNewSession = !this.remoteSessionIds.has(sessionId)` → `false` (still present), then looks up `reverseSessionIdMapping.get(sessionId)` → `undefined` → throws `No actual session ID found`.

The per-turn cleanup and the persistent-session teardown are conflated in one method.

## Reproduction (headless stdio)

1. Build the main process, launch `--headless --mode stdio` (seed `config.public.json` with `provider: ollama` + a `model` to pass the credential gate without a real key).
2. Send `{"type":"session.start","prompt":"hi"}`, read the `sessionId` from the event stream, wait for `session.end`.
3. Send `{"type":"session.message","sessionId":"<that id>","text":"again"}`.

Observed: `{"type":"error","message":"❌ An internal error occurred..."}` and the log line above. `session.abort` on the same id fails identically.

## Expected

A follow-up `session.message` / `session.abort` on an id the client received continues (or aborts) the existing session. Persistent session mappings should survive turn completion and only be removed on genuine session teardown (disconnect / explicit clear).

## Fix direction

Separate the two concerns: `clearSessionBuffer` (per-turn) should flush/clear only the ephemeral buffers, timers, and hashes — not the id mappings. Persistent mappings (`sessionIdMapping`, `reverseSessionIdMapping`, `remoteSessionIds`, channel/owner maps) should be torn down together in a dedicated session-end path so the two id structures never drift.

Introduced in #282.

lml2468:
Two related fixes in the `--headless --mode stdio` / `RemoteManager` path (both surfaced while verifying #282).

## 1. Multi-turn remote sessions (fixes #291)

Any follow-up message on an existing remote session failed with `No actual session ID found for remote session: remote-…` and the client got `❌ An internal error occurred.` Affected StdioChannel and every `RemoteManager`-backed channel; `session.abort` on an existing session failed too.

**Cause:** `clearSessionBuffer()` runs on every turn completion (`session.status` idle/error) to flush ephemeral per-turn buffers, but it also deleted the persistent id mappings (`sessionIdMapping`, `reverseSessionIdMapping`, channel/owner). `remoteSessionIds` is only ever added to, so turn 2 saw `isNewSession === false` yet couldn't resolve the actual id → throw.

**Fix:** Separate per-turn cleanup from teardown.
- `clearSessionBuffer()` clears only ephemeral state (buffers, timers, sent-hashes).
- New `removeRemoteSession()` tears down all persistent mappings in lockstep with `remoteSessionIds`; `clearRemoteSession()` resolves the actual id and calls it (now async; the IPC caller awaits).

## 2. Headless stdio process leak on stdin close

In stdio mode, `StdioChannel.handleClose()` logged "stdin closed" and emitted `!stop`, but nothing exited the process — it stayed alive indefinitely after the controller disconnected (had to be SIGTERM'd). RPC mode already exits on stdin close; stdio didn't.

**Fix:** Add an `onClose` callback to `StdioChannel` (invoked from `handleClose()`); the headless stdio entry point stops running sessions, runs `headlessCleanup()`, and `process.exit(0)` — mirroring RPC. Local re-entry guard prevents double shutdown.

## Tests

- `src/tests/remote/remote-manager-multiturn.test.ts` (new): two-turn flow continues (not re-starts, not throws); `clearSessionBuffer` preserves mappings; `removeRemoteSession` removes them. All three fail pre-fix, pass after.
- `src/tests/remote/stdio-channel.test.ts`: added a case asserting `onClose` fires on stdin close.
- Full `src/tests/remote` (25) + `src/tests/config` green; `tsc --noEmit` and eslint clean.

## Manual verification (real app, `--headless --mode stdio`)

Seeded `config.public.json` with `provider: ollama` + `model` to pass the credential gate without a real key.

- **Multi-turn** — turn-2 `session.message` before: `❌ internal error` + `No actual session ID found`. After: `Continuing session: … for remote: …` → `session.started` → `session.end`, no error.
- **Leak** — pipe one message then close stdin: before it hung until SIGTERM; after it self-exits `exit 0` in ~1s, logging `stdin closed → shutting down → Cleaning up`.

### DIFF
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -2879,9 +2879,9 @@ ipcMain.handle('remote.getRemoteSessions', () => {
   }
 });
 
-ipcMain.handle('remote.clearRemoteSession', (_event, sessionId: string) => {
+ipcMain.handle('remote.clearRemoteSession', async (_event, sessionId: string) => {
   try {
-    const success = remoteManager.clearRemoteSession(sessionId);
+    const success = await remoteManager.clearRemoteSession(sessionId);
     return { success };
   } catch (error) {
     logError('[Remote] Error clearing remote session:', error);
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -1152,7 +1152,8 @@ app
 
         // Shut down when stdin closes (controller disconnected), mirroring RPC
         // mode. Without this the process would stay alive indefinitely after the
-        // client goes away. Guard against re-entry via a local flag.
+        // client goes away. StdioChannel.handleClose() is idempotent so onClose
+        // fires once; the local flag is a cheap extra guard for the async body.
         let stdioClosing = false;
         stdioChannel.onClose(() => {
           if (stdioClosing) return;
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -1150,6 +1150,31 @@ app
 
         const stdioChannel = await remoteManager.startStdioMode(headlessArgs.cwd);
 
+        // Shut down when stdin closes (controller disconnected), mirroring RPC
+        // mode. Without this the process would stay alive indefinitely after the
+        // client goes away. Guard against re-entry via a local flag.
+        let stdioClosing = false;
+        stdioChannel.onClose(() => {
+          if (stdioClosing) return;
+          stdioClosing = true;
+          void (async () => {
+            log('[Headless] stdio stdin closed, shutting down');
+            if (sessionManager) {
+              for (const s of sessionManager.listSessions()) {
+                if (s.status === 'running') {
+                  try {
+                    await sessionManager.stopSession(s.id);
+                  } catch {
+                    // Best effort
+                  }
+                }
+              }
+            }
+            await headlessCleanup();
+            process.exit(0);
+          })();
+        });
+
         // Set the interceptor so ALL events from SM flow through stdio routing
         // (fixes the dead-code issue: SM calls headlessSendWithPermission directly,
         // which now checks stdioEventInterceptor before writing JSONL)
--- a/src/main/remote/channels/stdio-channel.ts
+++ b/src/main/remote/channels/stdio-channel.ts
@@ -63,6 +63,7 @@ export class StdioChannel extends ChannelBase {
   private rl: readline.Interface | null = null;
   private activeSessions: Set<string> = new Set();
   private closeHandler?: () => void;
+  private _closing = false;
 
   /**
    * Register a handler invoked when stdin closes (controller disconnected).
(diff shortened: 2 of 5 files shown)


lml2468:
Thanks for the review — addressed both points in 072fb6e:

1. **`handleClose()` idempotency** — added an internal `_closing` flag so `handleClose()` runs its cleanup (and fires `closeHandler` / the `!stop` abort emissions) exactly once, regardless of source (stdin natural close vs explicit `stop()`). The `index.ts` `stdioClosing` flag is now just a cheap extra guard for the async body. New test `invokes the onClose handler at most once (idempotent close)` asserts single invocation (verified it fails without the guard).

2. **Private `messageRouter` access in the test** — added a public `RemoteManager.routeMessage()` and routed stdio messages through it internally; the multi-turn test now calls the public API instead of reaching into the private field.

All `src/tests/remote` + `src/tests/config` green (167). Re-ran the real headless stdio app: multi-turn continuation and stdin-close self-exit both still work.

lml2468:
Closing this PR — the fixes (multi-turn session id mapping #291 + headless stdio stdin-close exit, with review refactor) have been merged and verified on my fork's `main` (lml2468/open-cowork@556e6c9). No longer pursuing the upstream merge via this PR.

PR #295, opened 2026-07-12
fix(chat): preserve per-session scroll position

### LINKED ISSUE #267: Chat scroll position resets when switching sessions or returning to session
### App Version

3.3.1

### Platform

Windows

### Bug Description

When switching between chat sessions or returning to a session, the scroll position in the chat view resets to the top (or jumps up about one page) instead of maintaining the user's last viewed position. This forces users to manually scroll down again to reach their most recent messages.

This is particularly frustrating during long conversations where users need to reference previous messages and then return to the current context.

### Steps to Reproduce

1. Open Open Cowork application
2. 2. Start a new chat session and send several messages to create a long conversation
3. 3. Scroll down to the bottom of the chat to see the latest messages
4. 4. Switch to a different session from the sidebar (or create a new session)
5. 5. Switch back to the original session
6. 6. Observe that the scroll position has reset to the top of the conversation
7. 7. User must manually scroll down again to reach the most recent messages
Alternative reproduction:
1. Have a long conversation in a session
2. 2. Scroll up to view older messages
3. 3. Click on the same session in the sidebar (without switching to another session)
4. 4. Observe that the scroll position jumps up instead of maintaining position
### Expected Behaviour

The scroll position should be preserved when:
- Switching between sessions
- - Returning to a previously viewed session
- - - Clicking on the same session in the sidebar
Ideally, the application should:
1. Save the scroll position (or message index) for each session when the user navigates away
2. 2. Restore the saved scroll position when the user returns to that session
3. 3. Default to the bottom of the conversation (latest messages) if no previous position was saved
4. 4. Maintain smooth scrolling behavior without jarring jumps
### Actual Behaviour

When switching between sessions or returning to a session:
- The scroll position jumps to the top of the conversation (or approximately one page up from the bottom)
- - The user's last viewed position is not saved
- - - Users must manually scroll down to reach their most recent messages
- - - - This behavior is consistent and reproducible
- - - - - The issue occurs regardless of how long the conversation is

```text

```

### Additional Context

**Impact:**
This issue significantly affects user experience, especially for users who:
- Work on long conversations and need to reference previous messages
- - Switch between multiple sessions frequently
- - - Use Open Cowork for extended periods
**Technical Details:**
The issue likely relates to how the chat view component handles session switching and state management. Possible causes:
- Session state is not persisting scroll position
- - Chat view is being unmounted/remounted on session switch without preserving scroll state
- - - No scroll position restoration logic is implemented
**Expected Implementation:**
A proper fix would involve:
- Storing scroll position (or message index) in session state
- - Restoring scroll position when a session is loaded
- - - Using `scrollIntoView()` or similar API to maintain position
- - - - Considering virtual scrolling optimizations for long conversations
**Workarounds:**
Currently, the only workaround is to manually scroll down after each session switch, which is tedious and disrupts workflow.

Sun-sunshine06:
## Summary

- keep scroll offsets in per-session UI state without rerendering message consumers
- restore each session before paint and default unseen sessions to the latest message
- clamp stale offsets safely when conversation content becomes shorter
- clear viewport state when sessions are deleted
- add focused store and scroll calculation regression tests

## Validation

- `npx vitest run tests/chat-scroll-position.test.ts src/tests/store/session-state.test.ts --reporter=verbose` (32 passed)
- `npm run typecheck`
- `npm run lint` (0 errors; existing warnings only)

Fixes #267

### DIFF
--- a/src/renderer/components/ChatView.tsx
+++ b/src/renderer/components/ChatView.tsx
@@ -64,6 +64,8 @@ export function ChatView() {
   const prevPartialLengthRef = useRef(0);
   const scrollTimeoutRef = useRef<ReturnType<typeof setTimeout> | null>(null);
   const scrollRequestRef = useRef<number | null>(null);
+  const scrollStateRequestRef = useRef<number | null>(null);
+  const pendingScrollStateRef = useRef<{ sessionId: string; scrollTop: number } | null>(null);
   const isScrollingRef = useRef(false);
 
   const hasActiveTurn = Boolean(activeTurn);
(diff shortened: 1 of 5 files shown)


Sun-sunshine06:
Addressed the review feedback in `75c4fb2`:

- scroll position persistence is now coalesced to at most one Zustand write per animation frame
- pending state is flushed safely during session switches/unmount
- normalized the inline comment to English

Validation remains green: 32 focused tests, typecheck, and lint with 0 errors.

PR #301, opened 2026-07-14
fix(ci): reset stale PR review context after history changes

Sun-sunshine06:
## Summary

- Reuse prior bot-review context only when the reviewed head is a verified ancestor on a complete, merge-free linear extension.
- Discard prior review IDs and run a fresh full review after rebase, force-push, rollback, merge commits, incomplete comparisons, or ancestry-check failures.
- Treat the current Files Changed list and current PR diff as authoritative in both Codex and DeepSeek review paths, with paginated PR file loading.
- Prevent an older matching review from suppressing a new run in an A -> B -> A history rollback.
- Clarify that bot reviews are advisory; a green workflow run only reflects automation completion.

## Why

PR #298 exposed the failure mode: the bot compared diverged heads (`04bcc817...e91a849`, 13 ahead / 7 behind) and inherited findings for files that were not in the PR's current seven-file diff. The workflow previously treated any different reviewed SHA as a valid follow-up.

The new checks fail closed: old review text and old-head compare data are available only for a verified safe linear update. All other cases reset review context and review the current PR from scratch.

## Validation

- `npx vitest run tests/deepseek-common.test.ts tests/codex-pr-review-context.test.ts` (11 tests passed)
- `npm run typecheck`
- `npm run lint` (0 errors; 8 existing warnings)
- `node --check` for both changed DeepSeek scripts
- Workflow YAML parse, Prettier check for tracked formatted file types, and `git diff --check`
- Live GitHub compare replay for PR #298: diverged range rejected; linear range accepted

The review event remains `COMMENT`; this PR does not turn advisory findings into a merge gate.

### DIFF
--- a/.github/prompts/codex-pr-review.md
+++ b/.github/prompts/codex-pr-review.md
@@ -39,9 +39,10 @@ Before any analysis, load PR metadata, latest head SHA, and diff from the GitHub
 Workflow-provided env:
 
 - `CURRENT_HEAD_SHA` - PR head SHA for this run
-- `LATEST_BOT_REVIEW_ID` - most recent prior bot review id, if any
-- `LATEST_BOT_REVIEW_COMMIT` - commit SHA reviewed by that prior bot review, if any
-- `IS_FOLLOW_UP_REVIEW` - `true` when contributor pushed new commits after the last bot review
+- `LATEST_BOT_REVIEW_ID` - latest reusable prior bot review id; empty when no prior context is safe to reuse
+- `LATEST_BOT_REVIEW_COMMIT` - commit SHA of that reusable prior review; empty when no prior context is safe to reuse
+- `IS_FOLLOW_UP_REVIEW` - `true` only when the prior reviewed head is a verified ancestor on a complete, merge-free linear extension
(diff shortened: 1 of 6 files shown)


Sun-sunshine06:
I verified the pagination findings against the real GitHub CLI/API before changing the implementation, and they are false positives.

Using this PR with `per_page=2` to force three pages:

```text
{"outerIsArray":true,"outerLength":3,"firstIsArray":true,"pageLengths":[2,2,2],"firstPath":".github/prompts/codex-pr-review.md"}
```

`gh help api` also states that `--slurp` wraps all pages of JSON arrays or objects in an outer JSON array. Running the committed helper directly against PR #301 returns all six changed paths, so the nested mock and `flatMap` are intentional and match real output.

The `gh pr diff --name-only` command is also intentional: this is an agent prompt, so its stdout is supplied directly to the reviewing model as the authoritative current Files Changed list; it does not need a shell variable to be consumed.

No code change is needed for these findings. The live commands used were:

```bash
gh api --paginate --slurp "repos/OpenCoworkAI/open-cowork/pulls/301/files?per_page=2"
node --input-type=module -e "import {listPullRequestFiles} from './.github/scripts/deepseek-common.mjs'; console.log(listPullRequestFiles('OpenCoworkAI/open-cowork', '301'))"
```

PR #326, opened 2026-08-09
fix(config): project contextWindow/maxTokens onto flat AppConfig

### LINKED ISSUE #262: Custom/Ollama contextWindow and maxTokens are not projected into runtime config
## Summary

The API settings UI lets users configure `contextWindow` and `maxTokens` for custom/Ollama providers. These values are preserved in provider profiles and even computed by `projectFromConfigSet()`, but they are not written back to the flat `AppConfig` returned by `configStore.getAll()`.

The agent runner consumes `configStore.getAll()` and reads the flat `runtimeConfig.contextWindow` / `runtimeConfig.maxTokens`, so user overrides can be ignored at runtime.

## Code path

Producer/UI:

- `src/renderer/types/index.ts:591-597` defines `ProviderProfile.contextWindow` and `ProviderProfile.maxTokens`.
- `src/renderer/types/index.ts:641-648` also defines flat `AppConfig.contextWindow` and `AppConfig.maxTokens`.
- `src/renderer/components/settings/SettingsAPI.tsx:344-385` renders Context Window and Max Tokens inputs for custom/Ollama providers.
- `src/renderer/hooks/useApiConfigState.ts:352-353` serializes these input values into persisted profiles.
- `src/renderer/hooks/useApiConfigState.ts:1736-1746` sends `profiles: persistedProfiles` in the `config.save` payload.

Store/normalizer:

- `src/main/config/config-store.ts:615-620` preserves valid numeric profile fields in `normalizeProfile()`.
- `src/main/config/config-store.ts:746-747` computes `contextWindow: activeProfile.contextWindow` and `maxTokens: activeProfile.maxTokens` in `projectFromConfigSet()`.
- `src/main/config/config-store.ts:957-981` builds the returned flat `AppConfig`, but does not include `projected.contextWindow` or `projected.maxTokens`.
- `src/main/config/config-store.ts:1008-1020` also composes projected config without those two fields.

Runtime consumer:

- `src/main/claude/agent-runner.ts:1591` reads `const runtimeConfig = configStore.getAll()`.
- `src/main/claude/agent-runner.ts:1630-1631` passes `runtimeConfig.contextWindow` and `runtimeConfig.maxTokens` into synthetic model resolution.
- `src/main/claude/agent-runner.ts:1652` queries Ollama `/api/show` when `runtimeConfig.contextWindow` is missing.
- `src/main/claude/pi-model-resolution.ts:158-159` falls back to known/default values when no override is passed.

## Steps to reproduce

Source-control-flow reproduction:

1. Configure a custom or Ollama provider.
2. Enter Context Window and Max Output Tokens in API settings.
3. Save the config.
4. Start an agent run with a model that depends on synthetic model resolution or Ollama context detection.

The numeric overrides are saved under profiles, but the runner reads the flat runtime config returned by `getAll()`, where these fields are omitted.

## Expected behavior

The active profile's `contextWindow` and `maxTokens` should be projected onto the flat runtime `AppConfig` consumed by the agent runner.

## Actual behavior

The values are preserved in profile data but dropped from the flat runtime projection, so the runner can fall back to default/registry/Ollama-probed values instead of the user's configured override.

## Suggested fix

- Add `contextWindow: projected.contextWindow` and `maxTokens: projected.maxTokens` to `normalizeConfig()`.
- Add the same fields to `composeProjectedConfig()`.
- Include `contextWindow` and `maxTokens` in `buildAgentRuntimeSignature()` so active sessions reload when only these numeric overrides change.

## Suggested tests

- Saving an active custom/Ollama profile with numeric overrides returns those values from `configStore.getAll()` as flat fields.
- Synthetic model resolution receives the saved `contextWindow` and `maxTokens` values.
- Runtime reload signature changes when only `contextWindow` or `maxTokens` changes.

---
Submitted with Codex.

AmirF194:
## Summary

`normalizeConfig()` and `composeProjectedConfig()` build the flat `AppConfig` object the
agent runner reads from `configStore.getAll()`. Both functions read every other projected
field from `projectFromConfigSet()`'s return value except `contextWindow` and `maxTokens`,
so a user's saved context-window/max-tokens override for a custom or Ollama provider profile
is silently dropped: the agent runner falls back to a default or auto-probed value instead of
what the user configured. `buildAgentRuntimeSignature()` (`src/main/index.ts`) has the same
gap, so even a user who edits only these two fields will not get the running session reloaded
with the new config, since the signature it hashes never changes.

Fixes #262.

## Fix

Project both fields through in `normalizeConfig()`, `composeProjectedConfig()`, and
`buildAgentRuntimeSignature()`.

`normalizeProfile()` (a few lines above `normalizeConfig()` in the same file) already
establishes the right pattern for this exact situation, with the comment "preserve optional
numeric fields so callers don't silently lose user-set values": assign the field only when it
is a valid positive number, never write it unconditionally. That matters here specifically
because `contextWindow`/`maxTokens` are genuinely absent on most profiles (no default profile
sets them), and `electron-store`'s underlying `conf` package throws on `.set()` if a key is
present with an explicit `undefined` value, unlike a plain object or `JSON.stringify`, which
both drop it silently. A first pass at this fix that assigned the fields unconditionally
(`contextWindow: projected.contextWindow`) crashed `ConfigStore`'s constructor for any profile
that never set them, which is the common case; `composeProjectedConfig()` additionally spreads
`...base`, so switching from a profile with an override to one without needs an explicit
`delete`, or the old value leaks into the new config set's projection.

## Testing

- `tests/config-store-context-window-projection.test.ts` (new): fails on `main` (both the
  drop itself and the constructor crash on a profile with no override), passes on this
  branch. Covers projection, persistence across a config-set switch, and clearing the value
  when switching to a set whose profile has no override.
- `tests/index-agent-runtime-signature.test.ts` (new): fails on `main`, passes on this branch.
- Full suite: `npx vitest run`, 154 files / 1108 tests pass. `npx tsc --noEmit` and
  `npx eslint src --ext .ts,.tsx` both pass with no new errors or warnings.
- Not verified: the `buildAgentRuntimeSignature()` gap as observed through a real running
  session reload (the fix is a one-line addition mirroring the six adjacent fields already
  handled the same way, but I did not drive the actual reload path end to end).


AmirF194:
No rush, just checking in after a week. CI is green and the branch is clean against main whenever you get a chance to look.

AmirF194:
Closing to keep the queue clean; happy to reopen if there's interest.

PR #331, opened 2026-08-16
fix(session): generate session titles in English

srbose:
﻿## Problem

Session titles in the chat list were generated in Chinese even for conversations held entirely in English. The title prompt in `session-title-utils.ts` was bilingual (English + Chinese instructions) and told the model to "reply in the same language as the user request" — in practice, models such as DeepSeek picked Chinese for English chats, so the sidebar filled up with Chinese titles while the rest of the UI stayed English.

## Fix

Make the title prompt English-only and always instruct the model to generate English titles, matching the app's UI.

## Notes

If multi-language titles are desired in the future, this could be extended to follow the app's i18n locale instead of being fixed to English. This PR intentionally keeps the change minimal.


srbose:
﻿Thanks for the review. To answer the open question: **forcing English titles is intentional for this PR.**

Rationale:

- The actual bug is that the bilingual prompt let the model pick a title language inconsistently — English conversations were getting Chinese titles while the rest of the UI stayed English. Making the instruction deterministic (English-only) fixes that directly.
- The reported case is exactly this: an English UI (locale `en`) showing Chinese sidebar titles, so the fixed behavior now matches the UI.
- Chinese-UI users lose native-language auto-titles, but titles remain editable by the user, and the chat *content* language is unaffected.

Your suggested locale-aware `buildTitlePrompt(prompt, locale)` is a good follow-up (e.g. pass the renderer's i18n locale into the title flow); I've noted it in the PR description as future work rather than expanding this change now.

The test additions were updated accordingly in the second commit.

PR #332, opened 2026-08-19
fix(remote): make permission approval work from Slack

srbose:
## Problem

Bash tool permission requests were effectively un-approvable from Slack. Four issues combined to make the flow fail:

1. **English replies were parsed as deny.** The reply parser accepted only `允许` / `y` / `yes` / `是` (and `始终允许` / `always`), so "allow", "approve", and "ok" all denied the request.
2. **Silent swallow on delivery failure.** `permission.request` events for remote sessions were consumed unconditionally; when remote handling returned `null` (missing session mapping, gateway down, send failure) the desktop dialog never appeared and the request silently auto-denied after 60s.
3. **The effective window was 60s, not 5 minutes.** The session-level 60s auto-deny timer ran independently of the remote flow's 5-minute timer, so users had at most a minute to reply.
4. **Thread-channel mismatch.** Slack reports threaded channels as `C123:threadTs`, so replies posted in the main channel never matched a prompt posted inside a thread (strict channel-id equality).

## Changes

- **`src/main/remote/interaction-utils.ts`** (new, dependency-free): `parsePermissionReply` accepts `y` / `yes` / `allow` / `approve` / `ok` / `1` (allow once), `always` (allow + remember), with Chinese tokens (`是` / `允许` / `始终允许`) kept for parity; anything unrecognized denies (fail-closed). `normalizeChannelId` strips Slack thread suffixes for matching.
- **`session-manager`**: permission timers are now deferrable — new `deferPermissionTimeout` clears and re-arms the auto-deny timer, and `handlePermissionResponse`/`requestPermission` settle through one idempotent path.
- **`remote-manager`**:
  - English Slack prompt listing the reply options (Feishu keeps its existing Chinese prompt).
  - After the prompt is delivered, the session timer is deferred to the same 5-minute window; expiry posts a notice to the channel before denying.
  - `handlePotentialInteractionResponse` matches in two passes: exact channel id first, then normalized (thread-suffix-stripped) comparison; owner check and consume-once semantics preserved.
  - `clearSessionBuffer` settles any pending permission as deny when a session ends, so its promise doesn't hang until the 5-minute timeout.
- **`index.ts`**: falls back to the desktop permission dialog when remote handling returns `null` or throws.

## Tests

44 new cases across three files:

- `tests/interaction-utils.test.ts` — reply parsing (all tokens, case/whitespace/punctuation variants, fail-closed) and channel-id normalization.
- `tests/remote-manager-permission-request.test.ts` — end-to-end permission flow: allow/always/deny replies, non-owner replies ignored, main-channel replies matching threaded prompts, send-failure cleanup, timer deferral, safe-tool auto-approve, session-end settlement.
- `tests/session-manager-permission-timeout.test.ts` — 60s default auto-deny, deferral to 5 minutes, reply-after-defer leaves no timer, `deferPermissionTimeout` returns `false` for settled/unknown ids.

Verified with `npx vitest run` on the three new files (44 passed), `npm run typecheck`, and `npm run lint` (0 errors).


srbose:
Addressed the review findings in 5f10719:

**[Major] Cross-thread matching** — normalized channel matching now only applies when at least one side is non-threaded (`(!isThreaded(channelInfo.channelId) || !isThreaded(channelId))`); exact matching already covers same-thread replies. A reply posted in a *different* Slack thread can no longer resolve a pending permission prompt. Tests updated: the cross-thread reply is asserted as not consumed (prompt still resolvable from the correct thread), plus a new case for a threaded reply to a prompt posted in the main channel.

**[Minor] 5-minute notice** — `deferPermissionTimeout` is now a required `AgentExecutor` member, so the advertised reply window is always backed by the session-level timer; the stdio executor gets an explicit no-op stub. The prompt also posts an expiry notice when the session timer has already fired (defer returns false).

**Testing** — the permission routing previously inlined in index.ts (with its null/throw fallback to the desktop dialog) is extracted into `src/main/remote/permission-route.ts`, with 7 new unit tests covering allow / allow+remember / deny / null fallback / thrown-error fallback / missing response handler.

Verification: 54 permission-related unit tests pass; `tsc --noEmit` clean; eslint 0 errors (8 pre-existing warnings in unrelated renderer files).

PR #334, opened 2026-09-02
fix(remote): keep session id mappings across turns (fixes #291)

### LINKED ISSUE #291: Multi-turn remote sessions fail: "No actual session ID found" after first turn (StdioChannel/RemoteManager)

haifengqiu:
## Summary

Every remote conversation (feishu / slack / stdio / websocket) dies after its
first agent turn: the follow-up message throws `No actual session ID found for
remote session: remote-…` and the user receives an internal-error reply.

Root cause (as analyzed in #291): `clearSessionBuffer()` runs on every turn
completion (`session.status` idle/error → `index.ts`) and deleted the
persistent session id mappings, while `remoteSessionIds` is append-only. On
turn 2, `isNewSession === false` but `reverseSessionIdMapping` has no entry → throw.

Fix — same direction acknowledged in #291 and proposed in the unmerged #292
(credit @lml2468), rebased onto current main:

- `clearSessionBuffer()` now clears ephemeral per-turn state only
 (response buffers, sent-message hashes, send timers)
- New `removeRemoteSession()` performs the full persistent teardown on genuine
 session end, keeping `remoteSessionIds` and the id mappings in lockstep
- `clearRemoteSession()` cascades into the persistent teardown (now async;
 IPC handler awaits)
- New public `RemoteManager.routeMessage()` — channels and tests route through
 RemoteManager instead of reaching into the private router
 (the review feedback from #292)

Follow-up messages now continue the same session via `continueSession`
(conversation context preserved across turns) instead of throwing.

Fixes #291. Supersedes #292.

## Type of change

- [x] Bug fix (`fix`)

## Checklist

- [x] `npm run test` passes locally (1104/1107 — see Testing note)
- [x] `npm run lint` passes locally (0 errors; 8 pre-existing warnings untouched)
- [x] UI changes tested on both macOS and Windows — N/A, no UI changes
- [x] New user-facing strings added to i18n files (`en` + `zh`) — N/A, no new strings

## Testing

New regression suite `tests/remote-session-mapping.test.ts` (4 cases, public
APIs only — drives messages via the new `routeMessage()` seam with a mocked
`AgentExecutor`):

1. Turn continuation: after `clearSessionBuffer()` (what `index.ts` calls on
  every turn completion), a second message on the same channel session calls
  `continueSession` instead of throwing or restarting.
2. Mapping preservation across per-turn cleanup.
3. `removeRemoteSession()` teardown: next message starts a fresh session
  (remoteSessionIds and mappings removed in lockstep — no drift).
4. `clearRemoteSession()` cascading teardown.

Reproduce the bug on main: send two consecutive messages to any remote
channel (e.g. headless `--mode stdio`, or feishu DM) — the second one fails
with `No actual session ID found for remote session`. Case 1 above covers
exactly this path and fails against main.

`npm run test`: 1104/1107 pass. The 3 failures are in
`src/tests/memory/*` (better-sqlite3 / fs environment issues on Windows);
verified they fail identically on `main` (ec5bd27) in this checkout — no
shared code paths with this PR. `npm run typecheck` clean; `npm run lint`
0 errors.

## Screenshots / recordings (if applicable)

N/A — no UI changes.


haifengqiu:
Thanks for the review — good catch on the id-kind contract.

**Caller audit**: the only caller today is the `remote.clearRemoteSession` IPC
handler (`src/main/index.ts`); the preload API documents the parameter as the
remote session id, and there is currently no renderer caller. So no existing
caller passes the wrong kind — but the footgun was real for future callers.

**Fix**: `clearRemoteSession()` is now id-agnostic (see new commit): it
normalizes either the router-side remote id (`remote-*`) or the actual agent
session id (UUID) before teardown, so passing the "wrong" kind can no longer
skip `removeRemoteSession()` and reintroduce the leak. The two id spaces
cannot collide.

**Tests added** (same commit):
- `clearRemoteSession` with the actual session id → full teardown
- `clearRemoteSession` with an unknown id → harmless no-op, existing session
 unaffected (continueSession still used for the next turn)

**On the pending-reply teardown test**: skipped deliberately — teardown reuses
the exact `clearSessionBuffer()` flush path (buffers + timers) that runs on
every turn completion; concurrent-flush semantics are unchanged by this PR.

PR #338, opened 2026-09-08
fix: decouple before-quit's guard from cleanupSandboxResources' own flag

### LINKED ISSUE #297: before-quit sets isCleaningUp before calling cleanupSandboxResources(), so quit cleanup is always skipped — and Cmd+Q needs two presses to exit
Summary
Two related defects in the macOS quit path (src/main/index.ts, present on current main and in v3.3.1):

Quit cleanup never actually runs. The before-quit handler sets isCleaningUp = true before calling cleanupSandboxResources() — whose first statement checks that same flag and returns immediately (logging [App] Cleanup already in progress, skipping...). Net effect: on every Cmd+Q, the entire cleanup body — remote-control stop, sandbox session sync-back to the host OS, sandbox/MCP shutdown, database close — is silently skipped.

Cmd+Q requires two presses to exit. The first quit is always cancelled via event.preventDefault() and then re-issued with app.quit() from inside the same async handler's continuation. In practice (observed on macOS, packaged build) the re-issued quit does not reliably terminate the app; a second Cmd+Q — which falls through the now-dead if (!isCleaningUp) guard without preventDefault — is what actually exits.
Observed behavior (packaged app, macOS)
Log excerpt from a real session (timestamps trimmed):

[App] Sandbox disabled, skipping bootstrap (using native mode)

[Lima] limactl found, checking instances...

[Lima] Instance not found in list

[App] Cleanup already in progress, skipping...   <- printed on FIRST Cmd+Q

After the skipping line the process remains alive; a second Cmd+Q exits it. Because cleanupSandboxResources() short-circuits, none of its steps run on quit — for users with active Lima/WSL sandbox sessions this skips the "sync changes back to host OS first" step, which is data-affecting, not just cosmetic.
Cause
// before-quit (current main)

isCleaningUp = true;          // set to prevent re-entrancy...

event.preventDefault();

await cleanupSandboxResources();  // ...but the callee guards on the SAME flag:

                                  //   if (isCleaningUp) { log('skipping'); return; }

app.quit();

The re-entrancy guard and the callee's own idempotency guard share one flag, so the guard defeats the work it was protecting.
Suggested fix (patch attached)
Introduce a quitReady latch: before-quit always preventDefault()s until cleanup has actually completed; a single shutdownAndQuit() helper runs cleanupSandboxResources() (which owns isCleaningUp itself, unchanged), bounded by a hard 60s withTimeout so a hung step can never strand the app, then sets quitReady and issues the one real app.quit().
Repeated Cmd+Q presses during cleanup land harmlessly (quit stays prevented; no duplicate cleanup).
window-all-closed (Windows/Linux and macOS dev) routes through the same helper, so both platforms share one shutdown path.
Handler is no longer an async listener relying on post-await re-issue mechanics; preventDefault() is decided synchronously.

Result: single Cmd+Q always runs full cleanup and exits; hang-proof via the existing per-step timeouts plus the overall cap.

npm run typecheck and eslint src/main/index.ts pass with the patch applied.
Environment
Open Cowork v3.3.x packaged build, macOS (Apple Silicon)
Reproduces with sandbox disabled (native mode), 0 MCP servers — i.e., the minimal path
Possibly related (separate issue?)
The packaged build also logs on startup: [AutoUpdater] Failed to load electron-updater: TypeError: Cannot read properties of undefined (reading 'checkForUpdatesAndNotify') — unrelated to the quit path, but it means affected builds won't self-update to a fixed release. Happy to file separately.

Proposed patch:
```
diff --git a/src/main/index.ts b/src/main/index.ts
index bdc24bc..39b7f98 100644
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -1473,6 +1473,24 @@ app
 // Flag to prevent double cleanup
 let isCleaningUp = false;
 
+// Set once cleanup has completed and the final quit is issued; lets that
+// quit pass through before-quit without being intercepted again.
+let quitReady = false;
+
+/**
+ * Run full resource cleanup (bounded by a hard timeout so a hung step can
+ * never strand the app), then issue the real quit exactly once.
+ */
+async function shutdownAndQuit(): Promise<void> {
+  try {
+    await withTimeout(cleanupSandboxResources(), 60000, 'Shutdown cleanup');
+  } catch (error) {
+    logError('[App] Shutdown cleanup failed, forcing quit:', error);
+  }
+  quitReady = true;
+  app.quit();
+}
+
 function withTimeout<T>(operation: Promise<T>, timeoutMs: number, label: string): Promise<T> {
   let timer: ReturnType<typeof setTimeout> | undefined;
   const timeoutPromise = new Promise<never>((_resolve, reject) => {
@@ -1573,8 +1591,7 @@ app.on('window-all-closed', async () => {
     // On Windows/Linux, closing all windows means quit.
     // On macOS dev mode, also quit — so vite-plugin-electron can restart cleanly
     // without the old process holding the single-instance lock.
-    await cleanupSandboxResources();
-    app.quit();
+    await shutdownAndQuit();
   }
   // On macOS production, keep app alive — cleanup happens in before-quit
 });
@@ -1585,30 +1602,30 @@ for (const sig of ['SIGTERM', 'SIGINT'] as const) {
 }
 
 // Handle app quit - before-quit (for macOS Cmd+Q and other quit methods)
-app.on('before-quit', async (event) => {
-  if (!isCleaningUp) {
-    // In dev mode, exit quickly — no need for async sandbox cleanup
-    if (process.env.VITE_DEV_SERVER_URL) {
-      stopNavServer();
-      try {
-        closeDatabase();
-      } catch {
-        /* best-effort */
-      }
-      closeLogFile();
-      tray?.destroy();
-      tray = null;
-      return;
-    }
-    // Set the flag immediately before any await to prevent re-entrant cleanup
-    isCleaningUp = true;
-    event.preventDefault();
+app.on('before-quit', (event) => {
+  // Cleanup has completed and shutdownAndQuit() issued this quit — let it through.
+  if (quitReady) return;
+
+  // In dev mode, exit quickly — no need for async sandbox cleanup
+  if (process.env.VITE_DEV_SERVER_URL) {
+    stopNavServer();
     try {
-      await cleanupSandboxResources();
-    } catch (error) {
-      logError('[App] before-quit cleanup failed, forcing quit:', error);
+      closeDatabase();
+    } catch {
+      /* best-effort */
     }
-    app.quit();
+    closeLogFile();
+    tray?.destroy();
+    tray = null;
+    return;
+  }
+
+  // Always hold the quit until cleanup has actually finished. Repeated Cmd+Q
+  // presses land here harmlessly while cleanup runs; shutdownAndQuit()
+  // re-issues the quit exactly once when done (or when its timeout fires).
+  event.preventDefault();
+  if (!isCleaningUp) {
+    void shutdownAndQuit();
   }
 });

AmirF194:
## Summary

Fixes #297.

`before-quit` set `isCleaningUp = true` right before calling `cleanupSandboxResources()`, whose first line returns early once that flag is already true. So the whole cleanup body (remote-control stop, sandbox sync-back, MCP shutdown, DB close) got skipped every time, and it took a second Cmd+Q to actually exit.

`cleanupSandboxResources()` already sets that flag itself, synchronously, right before its own first await, so before-quit doesn't need to touch it. Removed the duplicate set, and added a separate `quitReady` flag so a before-quit that fires while cleanup is still running (a second Cmd+Q, or the one `app.quit()` re-triggers once cleanup finishes) waits instead of letting the app exit mid-cleanup. `window-all-closed`'s own quit path sets `quitReady` too, for the same reason.

Targeted `main`, not `dev`: CONTRIBUTING.md says dev, but it's about 330 commits behind main and every recently merged PR I checked was based off main.

## Type of change

- [x] Bug fix (`fix`)

## Checklist

- [x] Commit messages follow Conventional Commits
- [x] Self-review completed, no debug logs, no commented-out code
- [x] `npm run test` passes locally
- [ ] UI changes tested on both macOS and Windows (main-process only, see Testing)
- [ ] New user-facing strings added to i18n files (none added)

## Testing

`src/main/index.ts` isn't importable in vitest (no exports, and the electron test mock's `app.on` is a no-op), so I followed the pattern already in `tests/index-config-window-behavior.test.ts` and asserted on the handler's source text instead. The new test fails on unpatched main (two `isCleaningUp` setters) and passes here; full suite, `tsc --noEmit`, and lint are all green too.

Couldn't exercise the real Electron quit sequence on this headless box, so the mid-cleanup deferral is reasoned from `before-quit`/`app.quit()`'s documented behavior and the code, not a live run.


Sun-sunshine06:
这个修复的代码逻辑和现有自动测试看起来没有明显问题，CI 也已通过；不过退出流程涉及 Electron 的真实事件时序，目前还缺少桌面环境下的人工验证，暂不建议只根据 source-scan 测试直接合入。

麻烦 @hqhq1025 帮忙重点验证以下场景：

1. Windows 正常关闭窗口/退出应用，确认应用能一次退出；
2. macOS 使用 `Cmd+Q` 退出，确认不再需要执行第二次；
3. 清理尚未结束时再次触发退出，确认不会提前退出，也不会重复执行清理；
4. 存在沙箱文件待同步时退出，确认同步正常完成；
5. Remote Control、MCP 和数据库资源能正常关闭；
6. 开发模式退出后，vite-plugin-electron 可以正常重启，不残留 single-instance lock。

如果以上场景通过，我认为这个 PR 可以合入。

AmirF194:
This has been open a week with only a passing bot check, no maintainer look yet. Flagging it in case it was missed.

AmirF194:
#354 fixes the same double-cleanup race, it even cites this issue in its own comment. This one was open twelve days earlier and is a smaller inline change; #354 pulls the logic into its own module with tests. Your call which one lands.

PR #342, opened 2026-09-13
feat: add optional you.com search MCP preset

mouse-value-add:
## What

Adds a `You` entry to the MCP connector quick-add presets (`MCP_SERVER_PRESETS` in `src/main/mcp/mcp-config-store.ts`), pointing at You.com's remote MCP server:

- **Preset URL:** `https://api.you.com/mcp?profile=free` — the keyless profile, so it works with a single click, no token, no env vars. The free profile exposes the `you-search` tool (basic web search).
- Because the preset requires no env vars, the existing `SettingsConnectors.tsx` flow adds it directly from "Quick Add Presets" — no UI changes were needed.
- Also added a short note in `readme.md` / `README_zh.md` explaining how to switch to the full authenticated toolset (`https://api.you.com/mcp` with an `Authorization: Bearer <YDC_API_KEY>` header, keys from https://you.com/platform/api-keys), which exposes `you-contents` (URL content extraction with citations) in addition to search.

## Why

Open Cowork agents currently have no built-in way to search the web — connectors cover browser automation and Notion, but nothing for plain "look this up" queries. A remote web search MCP connector fills that gap. I used the existing preset catalog mechanism rather than adding any custom code: the app already supports `streamable-http` remote servers with URL + headers, so this is a 8-line data entry plus tests.

## Opt-in behavior

- Nothing changes for existing users: the preset only appears in the preset list; a user must explicitly click Add, and the server is created disabled (`enabled: false`) until toggled on.
- No new dependencies, no default behavior change.

## Tests

Added `src/tests/mcp/mcp-config-store-presets.test.ts` (follows the existing `src/tests/mcp/` layout) covering:
- the preset exists and is `streamable-http` with the keyless URL
- no required env vars (so it's a one-click add)
- `createFromPreset('you-search')` produces a **disabled** server config
- unknown preset keys still return `null`

Validation run locally:
- `npx vitest run src/tests/mcp/` → 13/13 passing (my 4 new tests + the existing mcp-manager suite)
- `npm run typecheck` (tsc --noEmit) → clean
- Full `npm run test -- --run`: 1094 passing; the memory/sqlite test failures reproduce on a clean checkout of `main` in this Linux sandbox (better-sqlite3 native module built for Electron), so they're unrelated to this change.

## Notes

For anyone wanting You.com skills alongside the connector, the same endpoint family is packaged for other agent platforms (`npx skills add youdotcom-oss/agent-skills`) — not needed for this PR, just context on where the endpoint comes from.

Happy to adjust the preset name/key or drop the README notes if you'd prefer the smallest possible diff.

Closes #N/A (no upstream issue)


mouse-value-add:
Thanks for the thorough review — all three findings addressed in dc014f0:

- Removed the tautological `enabled` assertion from the preset-shape test (it could never fail given the preset type) — the disabled-by-default contract is now asserted solely via `createFromPreset`'s `enabled === false`, with a comment making that explicit.
- Moved the `requiresEnv` assertion onto the config produced by `createFromPreset('you-search')` so it would catch a regression where the factory drops or mutates `requiresEnv`.
- Added an inline comment on the `you-search` preset in `mcp-config-store.ts` noting it's keyless remote web search that stays disabled until the user opts in, and that queries egress to You.com only when enabled.

On the questions: I confirmed the Quick Add flow renders the `you-search` preset without a command or env prompt — the Connectors settings iterate the preset catalog and only show prompts for presets with non-empty `requiresEnv` (the software-development / gui-operate presets already exercise the empty-`requiresEnv` path, though they are stdio; the http branch skips the command prompt entirely). A UI-level test would be a nice follow-up but the store-layer assertions cover the data contract the UI reads.

Regarding surfacing connection/test status for the free-profile endpoint: that's really a feature of the connector framework rather than this preset (it would apply to any remote MCP server), so I'd rather keep this PR data-only and let that land as a separate improvement if the maintainers want it. Happy to sketch it out in an issue if useful.

mouse-value-add:
One correction to my previous comment: asserting `requiresEnv` on the config produced by `createFromPreset` doesn't typecheck — the factory's return type is `MCPServerConfig`, which doesn't carry `requiresEnv` (it's preset-catalog metadata, not server config). So in 715642c I kept the `requiresEnv` assertion on the catalog entry (where it guards the preset definition) and instead assert the produced config has no `command` and an empty `env` — that's the actual no-prompts contract for Quick Add on a keyless remote preset. `tsc --noEmit`, vitest, and eslint all pass.

mouse-value-add:
Addressed in 7c92fce:

- Both default-substituting assertions are now strict: `expect(preset.requiresEnv).toEqual([])` and `expect(config?.env).toEqual({})` — removing either field from the preset entry now fails the test.
- On the store round-trip question: `createFromPreset` spreads the preset into the saved config (`{ ...resolvedPreset, id, enabled }`), so I added a test that runs `saveServer` → `getServer` and asserts the persisted record stays `enabled: false` with the free-profile URL and empty `env` (cleanup via `deleteServer` in a `finally`). That covers the add-then-persist contract; the renderer-side Connectors UI add flow is precedented by the existing `gui-operate` preset path, so I didn't add a UI test for it.

All 14 tests in `src/tests/mcp` pass locally, eslint clean.

